PR #59, opened 2015-11-12
4.0.x sprint2 improvements

wwelling:


pom.xml
```diff
@@ -231,6 +241,14 @@
 				</plugin>
 			</plugins>
 		</pluginManagement>
+		<resources>
```
gadkrumholz: Are you sure this exclusion needs to go back in?

This was making all static files be excluded from war package for me.

src/main/java/org/tdl/vireo/AppContextInitializedHandler.java
```diff
@@ -24,11 +31,37 @@
...
+    
+    @Autowired
     ApplicationContext applicationContext;
     
+    final static Logger logger = LoggerFactory.getLogger(AppContextInitializedHandler.class);
+    
     @Override
     protected void before(ContextRefreshedEvent event) {
 
+    	// Why not configure context here?
```
gadkrumholz: I guess this code can go anywhere as long as it overrides internal property values and doesn't skip loading them instead. I did it as a static {} code in application.java just because it was easy and didn't require any annotations.

src/main/resources/static/ui/app/index.html
```diff
@@ -4,7 +4,7 @@
 <!--[if IE 8]>         <html lang="en" class="no-js lt-ie9"> <![endif]-->
 <!--[if gt IE 8]><!--> <html lang="en" class="no-js"> <!--<![endif]-->
 <head>
-	<base href="/vireo/">
+	<base href="/">
```
gadkrumholz: This is a perfect candidate for jsp-injected value... Are we sure we don't want to use jsp for the index.html instead?

PR #162, opened 2016-05-31
4.0.x sprint10 vir 176 142 jpa aggregation

wwelling:


wwelling:
Next steps will be to make workflow steps of organization and field profiles of workflow step be passed as entity instead of only ids. This will require removing annotations of server side and code changes on front end. Enhance field profile test and inheritance. Continue with workflows step update, such as setting to non overridable.

PR #330, opened 2016-12-08
Modified href paths for compatibility with Tomcat

cstarcher:
I modified the href paths in home.html to be compatible with Tomcat.


wwelling:
Can you make this pull request against 4.0.x-sprint19-staging? Thank you.

PR #368, opened 2017-02-16
4.0.x sprint21 field value test fix

wwelling:


wwelling:
Merged through fast forward.

PR #438, opened 2017-05-10
Allow org with same name under different parent org.

wwelling:


jcreel:
Amended constraint, changed tests accordingly.  Good reformatting on the UI.

PR #442, opened 2017-05-12
Fix RequestMapping for Tomcat compatibility

cstarcher:


jcreel:
Chris, could you test the Tomcat compatibility with the latest 4.0.x branch? After the latest Weaver upgrade, the request mapping should be handled without needing specific routes in the annotation.  If this works for you, we can close the PR.

cstarcher:
James, 4.0.x still does not work without these routes.

PR #571, opened 2017-08-31
Added route change listener to remove modal bg

jeremythuff:
This adds a listener to the route change success, in the runtime config, that closes the modal and removes the back drop on each route change.


jeremythuff:
This is being closed due to fixes in weaver

PR #577, opened 2017-08-31
4.0.x sprint26 vir 552 default filters

jeremythuff:
At user creation default filters are added to the user.

### DIFF
--- a/src/main/java/org/tdl/vireo/model/User.java
+++ b/src/main/java/org/tdl/vireo/model/User.java
@@ -93,7 +93,7 @@ public class User extends AbstractCoreUser {
     @OrderColumn
     private List<SubmissionListColumn> displayedSubmissionColumns;
 
-    @ManyToMany(cascade = { REFRESH, MERGE }, fetch = EAGER)
+    @ManyToMany(cascade = { REFRESH }, fetch = EAGER)
     @OrderColumn
     private List<SubmissionListColumn> filterColumns;
 
@@ -456,6 +456,19 @@ public class User extends AbstractCoreUser {
 
     public void setFilterColumns(List<SubmissionListColumn> filterColumns) {
         this.filterColumns = filterColumns;
+    }
+    
+    public void addFilterColumn(SubmissionListColumn submissionListColumn) {
+    	this.filterColumns.add(submissionListColumn);
+    }
+    
+    public void removeFilterColumn(SubmissionListColumn submissionListColumn) {
+    	this.filterColumns.forEach(fc->{
+    		if(fc.getId() == submissionListColumn.getId()) {
+    			int indexOfFc = this.filterColumns.indexOf(fc);
+    			this.filterColumns.remove(indexOfFc);
+    		}
+    	});
     }
 
     @Override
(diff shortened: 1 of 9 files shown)


src/main/java/org/tdl/vireo/model/User.java
```diff
@@ -457,6 +457,19 @@ public void loadActiveFilter(NamedSearchFilterGroup filter) {
...
+    
+    public void addFilterColumn(SubmissionListColumn submissionListColumn) {
+    	this.filterColumns.add(submissionListColumn);
+    }
+    
+    public void removeFilterColumn(SubmissionListColumn submissionListColumn) {
+    	this.filterColumns.forEach(fc->{
+    		if(fc.getId() == submissionListColumn.getId()) {
+    			int indexOfFc = this.filterColumns.indexOf(fc);
+    			this.filterColumns.remove(indexOfFc);
```
wwelling: Not safe to remove from a collection while iterating over it. Use a iterator and a while loop or set the item aside to remove or use a traditional for loop with a break.

src/main/java/org/tdl/vireo/model/repo/InputTypeRepo.java
```diff
@@ -6,6 +6,7 @@
 
 public interface InputTypeRepo extends JpaRepository<InputType, Long>, InputTypeRepoCustom {
 
+	// Name is not a unique constraint of input type, this method might fail
```
wwelling: Name most certainly should be unique. Should just make it so.

src/main/java/org/tdl/vireo/model/repo/impl/UserRepoImpl.java
```diff
@@ -30,8 +48,21 @@ public User create(String email, String firstName, String lastName, AppRole role
 
         newUser.putSetting("id", newUser.getId().toString());
         newUser.putSetting("displayName", newUser.getFirstName() + " " + newUser.getLastName());
-        newUser.putSetting("preferedEmail", newUser.getEmail());
+        newUser.putSetting("preferedEmail", newUser.getEmail());        
         newUser.setActiveFilter(activeFilter);
+        
+        List<SubmissionListColumn> defaultFilterColumns = null;
+		try {
```
wwelling: We now have a nice in memory default settings service. It should be used for this. I don't think the initial defaults are meant to be real time modified. Not sure user creation should have file I/O operations.

src/main/java/org/tdl/vireo/util/FileIOUtility.java
```diff
@@ -108,5 +115,16 @@ private Path processRelativePath(String relativePath) throws IOException {
         }
         return path;
     }
+    
+    public File getFileFromResource(String resourcePath) throws IOException {
```
wwelling: Probably don't need this in this utility. Not sure this file IO utility is a good utility in the first place.

src/main/java/org/tdl/vireo/util/FileIOUtility.java
```diff
@@ -28,6 +32,9 @@
 
     @Autowired
     private ObjectMapper objectMapper;
+    
+	@Autowired
```
wwelling: We might want to start considering minimizing autowiring singletons in multiple services. Don't have a exact reason as of now, but it seems careless to just autowire singletons everywhere you happen to need them for the moment and consider a more architectural approach. Just some thoughts.

src/main/java/org/tdl/vireo/model/repo/InputTypeRepo.java
wwelling: I looked and name is unique on InputType

PR #584, opened 2017-09-02
remove bower and ruby dependency, improved production build

wwelling:


cstarcher:
The btn name "Create a Name for Controlled Vocabulary" in controlledVocabularyManagement.html was requested by the product owner in VIR-504.

"Embargo" misspelled on lines 28 and 57 in embargoTypes.html

wwelling:
The button name "Create a Name for Controlled Vocabulary" is inaccurate. This needs to be discussed with the client. This is also a prelude to improving the UI somehow, with input from the client. Thanks for the review and misspelling findings. I am not entirely finished with this branch yet. Some grunt deploy issues. Works well until concat of JavaScript files.

cstarcher:
I agree. This is not the best wording. This issue was created at the same time as VIR-506, and I believe they are related. We should probably address them together.

PR #586, opened 2017-09-04
4.0.x sprint26 vir 275 org defaults merging

jeremythuff:


src/main/webapp/WEB-INF/app/views/admin/settings/organization/manageOrganization.html
```diff
@@ -68,7 +68,12 @@ <h3 class="panel-title">Manage {{getSelectedOrganization().name}}</h3>
 							    ng-if="getSelectedOrganization().id !== 1"
 								type="button"
 								class="btn btn-danger"
-								ng-click="openModal('#organizationConfirmDeleteModal')">Delete Organization</button>
+                ng-click="openModal('#organizationConfirmDeleteModal')">Delete Organization</button>
```
wwelling: Shouldn't the restore default workflow be under the Manage Workflow tab?

src/main/webapp/WEB-INF/app/views/modals/organization/organizationConfirmRestoreDefaultsModal.html
```diff
@@ -0,0 +1,17 @@
...
+  <button type="button" class="close" ng-click="cancelRestoreOrganizationDefaults()" aria-label="Close">
+    <span class="modal-close" aria-hidden="true">&times;</span>
+  </button>
+  <h3 class="modal-title">Resore Defaults</h3>
+</div>
+
+<form novalidate>
+  <div class="modal-body">
+    <h4>Really restore defaults for "{{ getSelectedOrganization().name }}"?</h4>
+    <p>This will result in the loss of all organizational customizations.</p>
```
wwelling: We probably want to indicate that all children organizations will have their workflow restored to default as well.

src/main/webapp/WEB-INF/app/views/modals/organization/organizationConfirmRestoreDefaultsModal.html
```diff
@@ -0,0 +1,17 @@
...
+  <h3 class="modal-title">Resore Defaults</h3>
+</div>
+
+<form novalidate>
+  <div class="modal-body">
+    <h4>Really restore defaults for "{{ getSelectedOrganization().name }}"?</h4>
+    <p>This will result in the loss of all organizational customizations.</p> 
+  </div>
+  <div class="modal-footer">
+    <button type="button" class="btn btn-default" ng-click="cancelRestoreOrganizationDefaults()">Cancel</button>
```
wwelling: Cancel does not close the modal.

src/main/webapp/WEB-INF/app/controllers/settings/organizationManagementController.js
```diff
@@ -61,6 +61,22 @@ vireo.controller("OrganizationManagementController", function ($controller, $q,
...
+				if(data.meta.type != 'INVALID') {
+					$scope.closeModal();
+					$timeout(function() {
+						AlertService.add(data.meta, 'organization/restore-defaults');
+					}, 300);
+				}
+			});
+		};
+
+		$scope.cancelDeleteOrganization = function() {
```
wwelling: This is a duplicate function name. Probably should run grunt to perform linting. This is also why the modal is not closing on cancel.

src/main/java/org/tdl/vireo/model/repo/impl/OrganizationRepoImpl.java
```diff
@@ -136,6 +138,21 @@ public void delete(Organization organization) {
 
         organizationRepo.delete(orgId);
     }
+    
+    @Override
+	public void restoreDefaults(Organization organization) {
```
wwelling: Not sure it is going to be this simple. I set up three organizations hierarchically. 

screen shot 2017-09-04 at 12 53 14 pm

I modified each child organization, thus creating new workflow steps as original on them.

I got the following errors trying to reset Test College. It did not succeed.

screen shot 2017-09-04 at 12 52 31 pm

I got the following errors trying to reset Test Department. It did not succeed.

screen shot 2017-09-04 at 12 53 04 pm

The Test Degree is the only one that was able to reset its workflow, but not sure if it cleaned up the workflow step it created as its original.

jeremythuff:
Test Added

src/main/webapp/WEB-INF/app/views/admin/settings/organization/manageOrganization.html
jcreel: Valid concerns, all.  These have been addressed with subsequent commits.

PR #634, opened 2017-10-31
moved identifier column to right of contacts column in controlled voc…

cstarcher:
…abulary management table in application settings


wwelling:
FYI this is vir-591

PR #648, opened 2017-11-08
Sprint27 email fixes

jcreel:


rladdusaw:
@jcreel What are all the cards implicated in this PR?

PR #649, opened 2017-11-08
Sprint27 vir 571 license files duplicated

rladdusaw:


jcreel:
The logic here clears out all existing licenses when the submission goes into "Submitted" status.  I don't think this is the desired behavior if the workflow is configured to let submitters upload additional/supplementary licenses.

jcreel:
We're going to put this logic in the code for now and make a card for more robust license upload and preservation when there's a well defined use case.

PR #662, opened 2017-11-27
Implement dynamic theming for Look And Feel configuration through custom WRO implementation

jsavell:


pom.xml
```diff
@@ -403,6 +403,9 @@
       <resource>
         <directory>src/main/resources</directory>
       </resource>
+      <resource>
+        <directory>src/main/webapp/WEB-INF/app</directory>
```
jsavell: This is added because the sass files located under src/main/resources/ reference sass files under app/node_modules, which is a symlink to the node_modules at the root dir of the repo.

Copying all these files to the target slows things down and we really only need a few of those files, but I couldn't come up with another way to make those node_modules sass files available to the src/main/resources/sass files during runtime.

PR #675, opened 2017-12-14
Sprint28 vir 627 reject

jsavell:


rladdusaw:
If you reject the application or embargo, and then later remove your rejection, the message that is sent to the action log is inaccurate. 

You see:
The committee has withdrawn its Application Approval.
The committee has withdrawn its Embargo Approval.

It should say something like:
The committee has removed its rejection of the Embargo.
The committee has removed its rejection of the Application.

PR #677, opened 2017-12-15
Sprint28 vir 636 access levels

rladdusaw:


jcreel:
This covers the bulk of the use cases.  However, need to add a new card to prevent Managers from affecting user privileges on Administrators.

PR #678, opened 2017-12-18
Changed all INPUT_TEXT fields with a CV to INPUT_SELECT

rladdusaw:


rladdusaw:
That was the advisor and it has been returned to an INPUT_CONTACT.

rladdusaw:
The subjects are being brought in through an Excel file in the /src/main/resources/proquest/ folder. It should still be a CV.

StephanieVUGSC:
Please hold this pull request until after beta testing.

PR #706, opened 2018-02-05
Converted Datetime values to timestamps

rladdusaw:


wwelling:
https://github.com/TexasDigitalLibrary/Vireo/blob/master/src/main/webapp/app/views/inputtype/input-datetime.html

This directive indicates it is possible to supply any type of date format. We need to know if this solution is able to convert any date time format to a timestamp. Also, is there a reason no to convert the date match to a timestamp to perform comparison?

wwelling:
I guess if we can confirm all persisted field values for input type date time are the same format, then this should do.

PR #816, opened 2018-02-27
organizations performance and submission list UX improvements

### LINKED ISSUE #724: During Organization creation, triptych for unrelated organizations pop up.
Each time I create a new organization, the triptych (2nd window) from the previous organization pops up.  Reselecting an organization will make it go away.   It is just confusing for anyone new to the system.

### LINKED ISSUE #786: Triptych not updating after organization changes
relates to #724

### LINKED ISSUE #814: Organization management performance
When creating several organizations the performance drops considerable on the organization management view.

This is due to the sheer size of the organization response. The more organizations that larger the response for requesting all organization.

The solution to this will be to have a partial view for all organizations and get the full organization upon selection.

wwelling:
resolves #814, resolves #724, resolves #786 

considerable performance gains on triptych

### DIFF
--- a/src/main/webapp/app/controllers/submission/submissionListController.js
+++ b/src/main/webapp/app/controllers/submission/submissionListController.js
@@ -451,6 +451,7 @@ vireo.controller("SubmissionListController", function (NgTableParams, $controlle
         $scope.resetColumns = function () {
             ManagerSubmissionListColumnRepo.reset();
             update();
+            $scope.closeModal();
         };
 
         $scope.resetColumnsToDefault = function () {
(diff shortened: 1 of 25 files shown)


src/main/webapp/app/controllers/submission/submissionListController.js
```diff
@@ -216,7 +214,7 @@ vireo.controller("SubmissionListController", function (NgTableParams, $controlle
 
         var resetBatchUpdateStatus = function () {
             $scope.advancedfeaturesBox.processing = false;
-            $scope.advancedfeaturesBox.assignee = findFirstAssignable();
```
jeremythuff: Originally the 0th user might not be "assignable", has that changed?

wwelling: this has changed. all the users in the list are now only assignable users.

PR #877, opened 2018-04-10
4.0.x 873 chair email visible

### LINKED ISSUE #873: Chair email address must be visible in Staff View without opening fields
Graduate staff frequently copy/paste the chair email address for communication.  It is not visible in the Staff  View in the tabs, or even if the field profile has been flagged.  Only the names show.  Email addresses are only visible when clicking on the pencil icon. 
image
image

cstarcher:
resolves #873

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2165,13 +2165,6 @@
             }
           }
         },
-        "string_decoder": {
-          "version": "1.0.1",
-          "bundled": true,
-          "requires": {
-            "safe-buffer": "5.0.1"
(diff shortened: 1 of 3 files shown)


rladdusaw:
The email is currently displaying as ["bobboring@maininator.com"] after the committee member's name. I think it would be better looking and easier to copy if we changed the value of lines 27 and 63 from:
{{field.value}} {{field.contacts}}
To:
{{field.value}}, {{field.contacts[0]}}

Issue #875, opened 2018-03-30
ORCID verification broken by API changes.


pobocks:
As of March 1, 2018, ORCID v1.2 is no longer supported.  As far as I can tell, the /orcid-bio route no longer exists, and definitely does not exist for unauthenticated clients of the public API.

The (substantially larger) response from the base URL appears to contain all necessary information, or /personal-details and /email combined do, though email address is not necessarily included in either due to privacy settings.

I don't have a working test environment, but my first step would be to replace the existing first line of the OrcidUtility class in `src/main/java/org/tdl/vireo/utility/OrcidUtility.java` with:
```java
private static final String ORCID_API = "http://pub.orcid.org/#";
```
and see if it works.

wwelling:
Let's confirm this is resolved.

wwelling:
This is working as expected. When ORCID validation is enabled, the ORCID entered during validation will make a request to ```https://pub.orcid.org/v2.0/#/``` and validate the ORCID against the user first name, last name, and email associated with the entered ORCID. One concern is that the validation is done against the vireo user and not first name and last name entered into the submission.

PR #881, opened 2018-04-24
Ref #875: ORCID API has changed, requiring updated validation code

pobocks:
The public bio ORCID response is no longer in a single namespace, is structured
differently, and is located at different URL.


rladdusaw:
Resolves #875

rladdusaw:
Resolves #875

PR #882, opened 2018-04-24
Ref #875: ORCID API has changed, requiring updated validation code

pobocks:
The public bio ORCID response is structured differently, and is located at different URL.


jcreel:
Resolves #875

StephanieVUGSC:
I have confirmed this fix. Approved.

PR #941, opened 2018-06-04
Add Return email recipients in Add File on submission

### LINKED ISSUE #904: Return email recipients in Add File
In Vireo 3, we are able to email recipients when using Add File to Replace a Primary Document.  This functionality has been removed in Vireo 4, and emailing is only allowed if Adding an Additional Document.   Replacing Primary Document with emailing is needed.  Please return Vireo 3 functionality.  See Vireo 3 screen, where email options are displayed by default.

image

kaladay:
Resolves #904 .


kaladay:
I am fine with removing the conditional on line 33.
Issue #904 specifically requests to solve the case for replacing existing primary documents.
It does also say "Please return Vireo 3 functionality.", so I can probably handle the case for new primary as well.

PR #880, opened 2018-04-20
879 now filtering by unassigned

cstarcher:
resolves #879

PR #942, opened 2018-06-05
June sprint #897 assigned to unassigned

### LINKED ISSUE #879: NullPointerException when removing Unassigned from Now Filtering By
java.lang.NullPointerException: null
	at org.tdl.vireo.controller.SubmissionListController.removeFilterCriterion(SubmissionListController.java:210) ~[classes/:na]

### LINKED ISSUE #897: Bug: Assigned To "Unassigned"
Staff Interface. List View. 

There are two problems occurring when "Unassigned" is selected in the ASSIGNED TO field.  First, the list is not filtered for submissions with no assignments.  Instead, all submissions disappear from view.  Second,  clicking on the X to the left of Unassigned under "Now Filtering By" does not remove it.  Normally that X will remove the term from the filter.

kaladay:
Resolves #897.
Resolves #879.

The issue is simply that a `WHERE IS NULL` check should be performed when unassigned.
The _furtherFilterByAssigneTo.html_ file is already passing `NULL` (until the solution provided by @cstarcher  here:  #880).

My approach is to utilize NULL as passed, which resolves both said issues.
My approach in conflict with #880.

An alternative approach to what I present here could be to depend on #880 as written, and then handle the case of where _"unassigned"_ is passed then perform a `WHERE IS NULL` check. The likelihood of somebody's name being _"unassigned"_ is close to 0, but I tend to prefer avoiding such situations altogether (however rare).

Feedback and discussion is desired.

### DIFF
--- a/src/main/java/org/tdl/vireo/controller/SubmissionListController.java
+++ b/src/main/java/org/tdl/vireo/controller/SubmissionListController.java
@@ -204,10 +204,20 @@ public class SubmissionListController {
     public ApiResponse removeFilterCriterion(@WeaverUser User user, @PathVariable String namedSearchFilterName, @RequestBody FilterCriterion filterCriterion) {
         NamedSearchFilterGroup activeFilter = user.getActiveFilter();
 
+        String filterValue = null;
         for (NamedSearchFilter namedSearchFilter : activeFilter.getNamedSearchFilters()) {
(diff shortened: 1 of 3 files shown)


kaladay:
I'm going to cherry-pick ce84a16 from #879 into this one.

PR #970, opened 2018-06-14
Return null if submission actionLogs are undefined or null

### LINKED ISSUE #969: TypeError: Cannot read property 'entry' of undefined for getLastActionEntry() and getLastActionDate()
The console log is flooded with messages such as:
```TypeError: Cannot read property 'entry' of undefined```

This happens on new submissions when they are first viewed in the admin view.

This causes the sidebox *Last action* and *What* fields to have text such as:
```{{box.getLastActionDate() | date : 'short'}}```
or 
```{{box.getLastActionEntry()}}```

This is also easily reproducible by accessing the add file modal to test PR #966.
Every time the (disabled) "*+*" action is selected for the e-mail, the console errors mentioned above appear.
This made it difficult to test and debug the other PR.

kaladay:
Resolves #969.

### DIFF
--- a/src/main/webapp/app/controllers/submission/adminSubmissionViewController.js
+++ b/src/main/webapp/app/controllers/submission/adminSubmissionViewController.js
@@ -362,11 +362,17 @@ vireo.controller("AdminSubmissionViewController", function ($anchorScroll, $cont
         };
 
         var getLastActionDate = function() {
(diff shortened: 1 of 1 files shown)


kaladay:
@cstarcher noticed that this is symptom of a bigger problem.
Action Logs should never be `null`.

The fixes currently in this PR would only hide/obscure the actual problem, so I am closing this PR in favor of a more proper solution.

PR #991, opened 2018-08-29
if field not optional validate on blur

### LINKED ISSUE #853: Display error message at point of error, not at Confirm & Submit
Please change Vireo 4 to behave like Vireo 3 in regard to error notices  (ie a required field was skipped, the wrong format was used, etc.)  Several testers noted that they would like this functionality returned to the way it was in Vireo 3 and not wait until the Confirm & Submit page to learn there are errors that need to be corrected.

wwelling:
Validation occurs on server side during the submission. If a field is modified it attempts to save onblur. Therefor, most cases the validation error would display. In the case of skipping a required was the only case not displaying a validation message when desired. Simple onblur will now attempt to save a required field whether it has been changed or not.

closes #853

### DIFF
--- a/src/main/webapp/app/directives/fieldProfileDirective.js
+++ b/src/main/webapp/app/directives/fieldProfileDirective.js
@@ -1,4 +1,4 @@
-vireo.directive("field", function ($controller, $filter, $q, $timeout, FieldValue, FileUploadService) {
+vireo.directive("field", function ($controller, $filter, $q, $timeout, FileUploadService) {
     return {
         templateUrl: 'views/directives/fieldProfile.html',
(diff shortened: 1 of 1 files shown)


src/main/webapp/app/directives/fieldProfileDirective.js
```diff
@@ -38,8 +38,9 @@ vireo.directive("field", function ($controller, $filter, $q, $timeout, FieldValu
             $scope.save = function (fieldValue) {
                 // give typeahead select time to save the value
                 $timeout(function () {
+                    console.log($scope.profile);
```
kaladay: Please remove console log.

PR #995, opened 2018-08-30
Provide error message and reset upload progress on file upload failures #855

### LINKED ISSUE #855: File upload problems in Student Interface
Stephanie was able to replicate some of the file upload problems in TAMU demo site 2:55, 3-19-18.  I was unable to upload a 1.2MB file as a Source or Administrative file.    After 30 seconds of waiting for the upload to complete, I moved to another tab.  I went back and the file was still uploading, but other beta testers reported that their file was missing when they went back to the screen.

kaladay:
There is already error messaging in place, but it is not implemented in every place.
Implement the errorMessage to be displayed in all of the file upload areas.

The error results from file upload failures is inconsistently structured.
Sometimes the errors are a string of HTML, other times the errors are objects.
This only parses the objects with known structures.

In all error cases, even if there is no parsed error message, an upload failure message is presented.

closes #855

### DIFF
--- a/src/main/webapp/app/controllers/submission/adminSubmissionViewController.js
+++ b/src/main/webapp/app/controllers/submission/adminSubmissionViewController.js
@@ -48,6 +48,8 @@ vireo.controller("AdminSubmissionViewController", function ($anchorScroll, $cont
 
     $scope.addCommentModal = {};
 
+    $scope.errorMessage = "";
+
     SubmissionRepo.findSubmissionById($routeParams.id).then(function(submission) {
 
         $scope.submission = submission;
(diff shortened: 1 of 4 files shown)


wwelling:
screen shot 2018-08-30 at 11 27 33 am

screen shot 2018-08-30 at 11 28 38 am

screen shot 2018-08-30 at 11 28 50 am

kaladay:
Those server failures where there before my changes and I think are a separate issue.
This is simply to provide the error message.
However, I never did see the `cannot set property of uploaded` error, so I will need to check to see if that is the a new issue introduced.

wwelling:
I am going to merge this and create an additional issue for the error to be handled immediately when server throws exception. Currently, if upload file exceeds max upload file size it waits until entire file is transferred to server before providing feedback on the front end.

PR #1019, opened 2018-09-05
publish location

### LINKED ISSUE #848: As a Manager, I would like to see the published location for a published submission

wwelling:
closes #848


wwelling:
It is fast forwarded from #1017 due to dependency and covers issue #761

PR #1071, opened 2018-10-08
Created MarcXML21 exporter

### LINKED ISSUE #713: MARC21XML export
MARC21XML  Export file will be needed for Vireo 4 as reported by VUG member. Not necessary for beta.

rladdusaw:
Resolves #713


src/main/java/org/tdl/vireo/controller/SubmissionController.java
```diff
@@ -577,7 +581,7 @@ private void processBatchExport(HttpServletResponse response, User user, String
                 response.setHeader("Content-Disposition", "inline; filename=" + packagerName + "." + packager.getFileExtension());
             } catch (Exception e) {
                 response.setContentType("application/json");
-
+                e.printStackTrace();
```
kaladay: Is this stacktrace intentional?

rladdusaw: It was just for logging. Removed.

kaladay:
It is working much better, but an on-line validator is throwing the following errors:
```    	
    The Prefix "xsi" For Attribute "xsi:schemaLocation" Associated With An Element Type "record" Is Not Bound., Line '1', Column '120'.
    The Prefix "xsi" For Attribute "xsi:schemaLocation" Associated With An Element Type "record" Is Not Bound.
```
I don't know the significance of that particular error.

kaladay:
By utilizing https://github.com/edsu/pymarc and yaz-marcdump (https://linux.die.net/man/1/yaz-marcdump) I was able to process and review the marc21 xml dump.

I think the export works fine.

PR #1075, opened 2018-10-11
Set body margin and padding to 0

### LINKED ISSUE #1070: Header not stretching width of the page

rladdusaw:
Resolves #1070


wwelling:
This appears to be a work around. Would be nice to know what is causing the problem. Also, why do they have to be !important. What is applying the margins to the body that can not be overridden by normal cascade?

kaladay:
I am unable to reproduce the bug to review this, but I have noticed that the problem *may* be caused by the "fade" class attached to the bootstrap modal. see: https://stackoverflow.com/questions/32862394/bootstrap-modals-keep-adding-padding-right-to-body-after-closed

rladdusaw:
@wwelling All I can see is it's being added to element.style so the !important is to make sure it's applied.

PR #1076, opened 2018-10-12
Reverted changes that were causing the submission to reload

### LINKED ISSUE #1067: After submission published to deposit location, submission refreshed multiple times

rladdusaw:
Resolves #1067.

This reverted changes made in #828 and #1011 that no longer seem to be needed.

My suspicion is that framework changes caused submission updates to be broadcast where we had been doing it manually in the past, causing multiple reloads.


wwelling:
Looks good, but would have to test before approving. Also, only partially reverting changes from both previous PRs.

PR #1116, opened 2019-01-14
4.0.x wro property shim

wwelling:


wwelling:
Do not merge. This will be resolved within Weaver. This shim is a work around and not the solution to the introduced regression.

Issue #1139, opened 2019-01-31
DSpaceMets export is broken


jsavell:
Trying to export to DSpaceMets results in:

UnsupportedFormatterException: Exporter does not support multiple docs !

This seems to be because the formatter pipeline was refactored to expect a Map, instead of a string, but this Exporter was unintentionally left out of the refactor.

PR #1141, opened 2019-01-31
[713] Restore MarcXML21 export

### LINKED ISSUE #713: MARC21XML export

jsavell:
Resolves #713 

This PR fixes the MarcXML21 export.

But the bug also affects the DSpaceMETS (#1139 ) and ProQuest (#1140 ) exporters.

The exports were failing silently because the UnsupportedFormatterException was being caught but not logged. The Exception is now being written to the log.

The catch block that was swallowing the UFException was also generating its own IllegalStateException because it was calling for a new response outputStream after the try block had  claimed it. The try and catch blocks are now sharing the same outputStream.

The result of this PR is that:
- MarcXML21 exporting works again, but **the resulting zip contents should be reviewed carefully to make sure they are still correct**.
- The DSpaceMets and ProQuest Export no longer fail silently, but still fail. Issues are open for each.
- MarcXML21, DSpaceMets, or ProQuest export failures will no longer trigger a follow on IllegalStateException

### DIFF
--- a/src/main/java/org/tdl/vireo/controller/SubmissionController.java
+++ b/src/main/java/org/tdl/vireo/controller/SubmissionController.java
@@ -21,6 +21,7 @@ import java.util.concurrent.ExecutionException;
 import java.util.zip.ZipEntry;
 import java.util.zip.ZipOutputStream;
 
+import javax.servlet.ServletOutputStream;
 import javax.servlet.http.HttpServletResponse;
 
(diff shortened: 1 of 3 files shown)


jsavell:
I generated a MarcXML21 export using a Vireo4 code base from before the bug was introduced.

I can confirm that the export generated by the code changes in this PR match the export from before the bug.

PR #1179, opened 2019-03-05
Feb sprint 1092 staff view email message

### LINKED ISSUE #1092: In staff, view an error notice is needed in Add Comment when a proper email is not provided.
I typed Student in the To field for email in Add Comment and clicked the plus sign.  "Student" disappeared and that was it.  I briefly saw a pop up or hover text about including an @ symbol.   Please add an error message if system will not accept what is input into the field.

jeremythuff:
Resolves #1092 : In staff, view an error notice is needed in Add Comment when a proper email is not provided.

This adds an alert error and error styling to the recipient and cc recipient inputs in the add comment modal. These will appear if you either blur or click add while the value contained within the input field is neither a correct email address nor the name of a contact.

*Out of scope*
This also adds a place holder text to each of the inputs, which was expressed as desirable during the last demo of this feature.


jeremythuff:
Good observations.

I have created an issue based on you suggestion regarding the addFileModal.

I will revisit the UX and see if I can come to an alternative solution that addresses your concerns.

jeremythuff:
Ok I have changed the UX. Blur now only runs a validation and never selects. The button is still in place and any attempt to click on it, while disabled will run the validation on the field. This functionality feels more like the current UX, only with the addition of the validation message.

PR #1187, opened 2019-03-08
fixed assignee update in submission status sidebox

### LINKED ISSUE #1161: Bug - Assignee will not populate until page is refreshed
In List View, the value in the Assign To/Assignee field in the Submission Status box does not update until the page is refreshed.  Note that the action is recorded in the Log without a page refresh. 

image

cstarcher:
resolves #1161


src/main/webapp/app/controllers/submission/adminSubmissionViewController.js
```diff
@@ -53,6 +53,8 @@ vireo.controller("AdminSubmissionViewController", function ($anchorScroll, $cont
         });
         // update current submissions status
         angular.extend($scope.submission.submissionStatus, submission.submissionStatus);
+        // update current assignee
```
jcreel: Could try:
`
angular.extend($scope.submission, { assignee : submission.assignee });
`

PR #1207, opened 2019-03-15
Feb sprint mets export update

smutniak:
Updates METS export to work with multiple records and present them in a structure aligned with that produced by vireo3.


smutniak:
superceded by another branch

PR #1241, opened 2019-05-02
4.0.x

jeremythuff:


smutniak:
The only thing I noticed amiss was changes to lines 28-29 and 41-42 in vireo/model/Submission.java These are likely superfluous residue from something I was trying to fix but was given a better approach by William Welling

PR #1256, opened 2019-05-16
prevent reorder when dropped outside of list

### LINKED ISSUE #1223: Error Message : Destination is out of range
The error message "Destination is out of range" appears when re-ordering email templates.  Since the SYSTEM templates cannot be edited, institutions will prefer to order their templates so that the SYSTEM ones are the last in view.  

Reordering works, although the numbers do not change, so I am not sure what the error is. 
image

wwelling:
resolves #1223

### DIFF
--- a/src/main/webapp/app/controllers/settings/graduationMonthRepoController.js
+++ b/src/main/webapp/app/controllers/settings/graduationMonthRepoController.js
@@ -91,7 +91,6 @@ vireo.controller("GraduationMonthRepoController", function ($controller, $scope,
 		};
 
 		$scope.removeGraduationMonth = function() {
-			$scope.modalData.month = getMonthIndex($scope.modalData.month);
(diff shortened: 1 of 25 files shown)


wwelling:
- corrected issue above when dragging a newly created entity
- corrected create new after a drag has occured

src/main/webapp/app/controllers/settings/graduationMonthRepoController.js
```diff
@@ -20,6 +20,10 @@ vireo.controller("GraduationMonthRepoController", function ($controller, $scope,
 	    'July', 'August', 'September', 'October', 'November', 'December'
 	];
 
+	var getMonthIndex = function(month) {
+		return months.indexOf(month);
+	}
```
kaladay: missing semicolon.

Issue #1127, opened 2019-01-25
initial package build of Vireo 4.x doesn't copy application.yaml into war file


bricas:
I did the build as per the instructions:

`
mvn clean package -DskipTests -Dproduction -Dassets.uri=file:/opt/vireo/ -Dconfig.uri=file:/opt/vireo/config/
`

however, loading up the war file (with java -jar), it complains about $production$ not being a boolean.

I was able to copy `target/classes/application.yaml` to `src/main/resources/application.yaml`, rebuild, and have a working .war file.

StephanieVUGSC:
related to #840

PR #1281, opened 2019-06-05
jar and war deploy delimiter fix in pom.xml and application.xml - als…

smutniak:
…o fixed resource directory location in service

pom.xml and application.xml changes:
war file build ignores the pom.xml delimiter of ('$*$') and uses default ('${*}') instead.
jar file adheres to delimiter directive.
Set application.yaml to use the default so only need to change the package elements value to 'jar' or 'war'.

service changes:
Fixed spelling error in AssetService.java and SystemDataLoader.java for 'Resouce'
Also fixed detection of jar vs war deploy for finding resource directory which is packaged in different locations based on jar vs war.
Scheme detection was not sufficient as both were invoked using 'jar' but the full URI string can be used to find the suffix of .jar or .war.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -374,7 +374,7 @@
         <artifactId>maven-resources-plugin</artifactId>
         <configuration>
           <delimiters>
-            <delimiter>$</delimiter>
+            <delimiter>${*}</delimiter>
           </delimiters>
         </configuration>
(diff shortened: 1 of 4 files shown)


smutniak:
Fixes #1127

PR #1285, opened 2019-06-11
Sprint posttcdl 1231 batch email advisors

### LINKED ISSUE #1231: Add Vireo 3 functionality to batch email Advisors
Emailing Advisors in batch is a critical function for Vireo users.   Please hard-code an option for Advisor that would look for any field profile with the "dc.contributor.advisor" predicate.  

 Users will be instructed not to change the predicate or make the field unavailable in a workflow, with warning that otherwise the batch email to advisor will not work. 

In keeping with Vireo 3 functionality, we will also instruct users to not make this field repeatable so that only 1 email address is allowed.

smutniak:
Add 'Advisor' option to batch emails.  This will add the email address associated with dc.contributor.advisor value to the list of email recipients when 'Advisor' is chosen.
Fixes #1231


smutniak:
want to test further with 1272 fix

PR #1290, opened 2019-06-17
Sprint posttcdl staging

jcreel:


smutniak:
It contains all and only those changes previously approved so I'll merge.

PR #1293, opened 2019-06-18
E-mail forms (comment and comment/batch) should only add e-mail once

### LINKED ISSUE #1273: Feature Request: E-mail forms (comment and comment/batch forms) should only add e-mail once
Sending e-mails via the "comment" or even the "batch/comment".
It would be cleaner to have the e-mail specified only one time.

For a single comment, when specifying, I can manually add say "test@example.com" two times to the form and then add "Submitter" (assuming submitter has email of "test@example.com").
The e-mail **test@example.com** would appear 3 times in the **To** E-mail Header.

I believe it would be cleaner and potentially less error prone to:
1. Prevent manually adding the same e-mail twice.
2. When populating the e-mail To/CC headers, do not add the same e-mail more than once.

Extra long or repetitive e-mails addresses are potential grounds for being flagged as spam. Especially if a batch e-mail is sent for numerous submissions.

A generated to header of
`To: <test@example.com>`
is much cleaner and less problem prone than:
`To: <test@example.com>, <test@example.com>, <test@example.com>`

kaladay:
Moves the submission email sending to its own file as a Service.
Adds additional logic to ensure that if an e-mail is already present in the TO or CC e-mail headers, do not add the duplicate e-mail.
Very basic tests are implemented.

Resolves #1273

### DIFF
--- a/src/main/java/org/tdl/vireo/service/SubmissionEmailService.java
+++ b/src/main/java/org/tdl/vireo/service/SubmissionEmailService.java
@@ -220,9 +220,8 @@ public class SubmissionEmailService {
         List<String> recipients = new ArrayList<>();
         List<HashMap<String, Object>> emails = (List<HashMap<String, Object>>) data.get(property);
 
-        emails.forEach(emailRecipientNode -> {
-            String type = (String) emailRecipientNode.get("type");
-            EmailRecipient recipient = buildEmailRecipient(type, emailRecipientNode, submission);
+        emails.forEach(emailRecipientMap -> {
+            EmailRecipient recipient = buildEmailRecipient(emailRecipientMap, submission);
             if (recipient != null) {
                 for (String recipientEmail : recipient.getEmails(submission)) {
                     if (!recipients.contains(recipientEmail)) {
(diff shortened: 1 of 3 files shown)


src/main/java/org/tdl/vireo/service/SubmissionEmailService.java
```diff
@@ -0,0 +1,288 @@
...
+    public void sendAdvisorEmails(User user, Submission submission) {
+        InputType contactInputType = inputTypeRepo.findByName("INPUT_CONTACT");
+        EmailTemplate template = emailTemplateRepo.findByNameAndSystemRequired("SYSTEM Advisor Review Request", true);
+
+        String subject = templateUtility.compileString(template.getSubject(), submission);
+        String content = templateUtility.compileTemplate(template, submission);
+
+        List<String> fullRecipientList = new ArrayList<>();
+
+        // TODO: this needs to only send email to the advisor not any field value that is contact type
```
rladdusaw: I can see that this was carried over from the original implementation, but I think it might be the time to fix it. We don't want the entire committee getting the adviser approval link or anything like that.

@jeremythuff @wwelling @jsavell @jcreel Thoughts on this?

jcreel: This needs a fix soon.  Let's resolve this question and address the problem in the context of issues of #1303 and #1305 .

Issue #1297, opened 2019-06-19
Marc21 and DSpace exports do not properly render advisor/chair when field is repeatable


smutniak:
marc21 export should provide multiple '720' records when there is a co chair
dspace Simple Archive Format also needs to render multiple chairs in dc.contributor.advisor

PR #1298, opened 2019-06-19
Sprint july release fix exporters to render multiple advisors

smutniak:
Fixes recently discovered issue #1297


smutniak:
Also fixes remaining dateTimeFormat problem encountered upon proquest export,  already fixed for methods used by other exporters

smutniak:
@rladdusaw Proquest has been broken for a while as expressed in issue 1140 which is saved for a later release.  It shouldn't be hard based on other work I did but I can't get to it yet. This pr is just for allowing multiple advisors for those exports that do work or when they work.  Stephanie specifically requested this with Dspace SAF in mind.

Issue #1301, opened 2019-06-20
dspace export format prefixes in contents file cause import problems, suffixes not needed


smutniak:

rladdusaw:
@smutniak Were these prefixes and suffixes needed for older versions of DSpace? We might want to consider leaving a template file for DSpace 5 exports if backwards compatibility is a concern.

PR #1302, opened 2019-06-20
removed prefixes meant for another format and suffixes except primary…

smutniak:
…:true for dspace export contents file

fixed export 'contents' file so dspace zip file can now be successfully imported by dspace6.

### DIFF
--- a/src/main/java/org/tdl/vireo/controller/SubmissionController.java
+++ b/src/main/java/org/tdl/vireo/controller/SubmissionController.java
@@ -732,7 +732,7 @@ public class SubmissionController {
                     if (exportPackage.isMap()) {
                         for (Map.Entry<String, File> fileEntry : ((Map<String, File>) exportPackage.getPayload()).entrySet()) {
                             zos.putNextEntry(new ZipEntry(submissionName + fileEntry.getKey()));
-                            contentsText.append("MD " + fileEntry.getKey() + "\n");
+                            contentsText.append(fileEntry.getKey()+"\n");
                             zos.write(Files.readAllBytes(fileEntry.getValue().toPath()));
                             zos.closeEntry();
                         }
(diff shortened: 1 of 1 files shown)


smutniak:
fixes #1301

src/main/java/org/tdl/vireo/controller/SubmissionController.java
```diff
@@ -743,7 +743,7 @@ private void processBatchExport(HttpServletResponse response, User user, String
                         Path path = assetService.getAssetsAbsolutePath(ldfv.getValue());
                         byte[] fileBytes = Files.readAllBytes(path);
                         zos.putNextEntry(new ZipEntry(submissionName + ldfv.getFileName()));
-                        contentsText.append(ldfv.getFileName() + " bundle:LICENSE\n");
+                        contentsText.append(ldfv.getFileName()+"\n");
```
kaladay: I believe `bundle:LICENSE` (and `bundle:CONTENT` below) are supported and used by Dspace 6:

I think these two should still be supported as they are specific DSpace import helpers.

Is there some specific reason as to why these two have been removed?

see: https://wiki.duraspace.org/display/DSDOC6x/Importing+and+Exporting+Items+via+Simple+Archive+Format#ImportingandExportingItemsviaSimpleArchiveFormat-DSpaceSimpleArchiveFormat

smutniak: Since they are optional the task was deferred and set as issue 1313 pending another conversation with Stephanie.

smutniak: I could put in bundle:ORIGINAL and bundle:LICENSE but wasnt sure what others could be added.  I've retained primary:true for the primary doc.    I need to run now but can add the above in a separate PR on monday.  Other bundle suffixes could be added later.

smutniak: PR branch updated

PR #1304, opened 2019-06-24
Sprint july release 840 update readme

rladdusaw:


smutniak:
Please fix spelling errors on lines:
301,315,318,320,470

Since we mention running as spring-boot:run for development purposes should we also include the
generate and accounts commands that one can give when the -Drun.arguments=console flag is used?

PR #1308, opened 2019-06-25
Sprint july release 1273 email once andthen1231

### LINKED ISSUE #1231: Add Vireo 3 functionality to batch email Advisors

smutniak:
add advisor to batch email on top of kevin day's earlier fixes
fixes #1231


smutniak:
@rladdusaw karma test is fixed
I'll soon be following this PR with another one which builds upon it.  that should be the last one

PR #1315, opened 2019-07-01
update accounts generation to separate student/reviewer/manager from …

smutniak:
…admin

test-etd.tdl.org will provide test accounts for users who with to quickly set up to test drive features held by different account roles.  The changes in this PR separates admin test account generation from the creation of student, reviewer, and manager accounts.  This will allow me to set up the latter 3 accounts but not provide admin test accounts on the public test-etd.tdl site.

### DIFF
--- a/src/main/java/org/tdl/vireo/cli/Cli.java
+++ b/src/main/java/org/tdl/vireo/cli/Cli.java
@@ -126,16 +126,32 @@ public class Cli implements CommandLineRunner {
                     }
                     for (int i = 0; i < acct; i++) {
                         String enc_pwd = passwordEncoder.encode("password");
-                        User submitter = userRepo.create("anonymous" + (i + 1) + "@example.com", "anonymous", "example", enc_pwd, Role.ROLE_ANONYMOUS);
-                        userRepo.saveAndFlush(submitter);
-                        submitter = userRepo.create("student" + (i + 1) + "@example.com", "student", "example", enc_pwd, Role.ROLE_STUDENT);
-                        userRepo.saveAndFlush(submitter);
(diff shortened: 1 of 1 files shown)


src/main/java/org/tdl/vireo/cli/Cli.java
```diff
@@ -126,16 +126,32 @@ public void run(String... arg0) throws Exception {
...
-                        submitter = userRepo.create("student" + (i + 1) + "@example.com", "student", "example", enc_pwd, Role.ROLE_STUDENT);
-                        userRepo.saveAndFlush(submitter);
-                        submitter = userRepo.create("reviewer" + (i + 1) + "@example.com", "reviewer", "example", enc_pwd, Role.ROLE_REVIEWER);
-                        userRepo.saveAndFlush(submitter);
-                        submitter = userRepo.create("manager" + (i + 1) + "@example.com", "", "manager", enc_pwd, Role.ROLE_MANAGER);
-                        userRepo.saveAndFlush(submitter);
-                        submitter = userRepo.create("admin" + (i + 1) + "@example.com", "", "admin", enc_pwd, Role.ROLE_ADMIN);
-                        userRepo.saveAndFlush(submitter);
+                        User testacct = userRepo.create("student" + (i + 1) + "@example.com", "student"+(i+1), "example", enc_pwd, Role.ROLE_STUDENT);
+                        System.out.println("Creating account with email "+testacct.getEmail()+" with role ROLE_STUDENT");
```
rladdusaw: We should be using the [logger](https://github.com/TexasDigitalLibrary/Vireo/blob/master/src/main/java/org/tdl/vireo/controller/DegreeController.java#L30) instead of System.out calls.

Issue #712, opened 2018-02-08
MARC21 export


jcreel:
MARC21 Export file will be needed for Vireo 4 as reported by VUG member.  Not necessary for beta, but would be helpful.

StephanieVUGSC:
file type should be .mrc  and not .bib

PR #1359, opened 2019-08-30
[712] marc21 exporter

kaladay:
resolves: #712


kaladay:
Will needs some minor tweaks.
The UTF-8 does work, except the MARC record calculated size is calculated by characters and needs to be by Bytes to accomodate the different byte-width sizes in UTF-8. Other than the size calculation issue, I was able to confirm that the MARC records generated by this (both ASCII and UTF-8) work.

PR #1371, opened 2019-09-09
August sprint staging

smutniak:


smutniak:
In files such as src/main/webapp/app/views/directives/dragAndDropEmbargo.html th
ere is a change from isSystemRequired to systemRequired.  There are are similar
changes in various java classes and .json init files.  
  
The method javascript method isSystemRequired(modalData) is introduced in
 src/main/webapp/app/views/modals/settings/embargoEditModal.html
Is that intended?

isSystemRequired also appears in  src/main/webapp/tests/unit/controllers/settings/emailTemplateRepoControllerTest.js and  src/main/webapp/tests/unit/controllers/settings/embargoRepoControllerTest.js

kaladay:
The change from `isSystemRequired` into `systemRequired` was a necessary and intended change.
see: https://github.com/TexasDigitalLibrary/Vireo/pull/1345

There may be some UI functions that define `isSystemRequired` but it does some with different meaning. That is to say, it is not directly referenced a database entity `isSystemRequired`, it is instead referring to if this "is" a "systemRequired" field.

I apoligize if this is confusing.
`embargoRepoController.js` can be used as an example for where a method `isSystemRequired()` is used to check if `systemRequired` model property is TRUE. This could be potentially be renamed to just `systemRequired()`.

PR #1376, opened 2019-09-23
August sprint staging 1140 proquest

smutniak:
This fixes the proquest export.  Since it was a late fix it was based off of august-sprint-staging.


smutniak:
I have hosted this branch on test-etd and the proquest output was tested and verified by stephanie.  this fixes #1140

PR #1401, opened 2019-11-26
fix thesis.degree.college||school||program per 1383

smutniak:
fixes 1383


smutniak:
moving to another branch

PR #1403, opened 2019-11-26
4.0.x v1fixes filesize limitexceeded 1396

smutniak:
experimental fix to 1396 - takes a while to replicate


smutniak:
wrong base

PR #1406, opened 2019-11-26
fix thesis.degree.college||school||program per 1383

smutniak:


smutniak:
merging into _staging_migration branch just for my own migration work

PR #1412, opened 2020-01-23
Dependabot/maven/org.apache.tika tika core 1.22

smutniak:


smutniak:
only updating apache.tika dependency in master for now

Issue #1424, opened 2020-02-14
Graduation Semester Filter gets stuck if the intermodal selectors year banner is clicked


smutniak:
clicking on the year banner changes the selection from current year and graduation months to a year selector.  No selections on this are permitted to return to a year/month selection so it dead ends unless a page refresh is performed.  Discovered by Erika at TTU.

PR #1437, opened 2020-04-15
4.0.x april2020 1424

mtpatterson:
Resolves issue #1424 

Disables the year view entirely, but allows users to navigate years with the arrow buttons.


smutniak:
Looks good. Thank you!  Note: this eliminates the year picker panel and requires the user to just use the earlier/later year increment selector therefore avoiding the part which caused the dead end in the intermodal.

Issue #1434, opened 2020-03-27
TinyMCE 5.1.6 causing issues


ublrama:
With the update of TinyMCE in vireo, it is not possible to edit text boxes.

![image](https://user-images.githubusercontent.com/50174699/77764884-ae6e4680-703d-11ea-8839-faea14c1872d.png)

In TinyMCE 5.1.6 there is no folder named "modern".

PR #1439, opened 2020-04-17
Resolve issues with upgrading TinyMCE from 4.x to 5.x

nwoodward:
The TinyMCE major version update from dependabot (https://github.com/TexasDigitalLibrary/Vireo/commit/7154baec90cff4eae10b4ad5590288c4aaf9de08) created a few issues with TinyMCE textareas. As noted in #1434, the Modern theme is no longer available in version 5.x. Additionally, the toolbar attribute has a different format, and the Autoresize plugin is now required to expand the textarea to fit the <div>.

NOTE: While debugging this issue, I noticed that $scope.plugins is undefined in the lockingtextarea directive. If that should be set somewhere else or in a different fashion, please let me know. 

References:
TinyMCE 4 to 5 migration guide: https://www.tiny.cloud/docs/migration-from-4x/#removededitorsettings

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "ng-csv": "0.3.6",
     "ng-sortable": "1.3.8",
     "ng-table": "3.0.1",
-    "tinymce": "4.9.2"
+    "tinymce": "5.2.1"
   },
   "devDependencies": {
     "grunt": "1.0.3",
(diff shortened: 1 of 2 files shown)


nwoodward:
@ublrama could you test this PR to see if it resolves your issue?

Issue #1131, opened 2019-01-28
Handle request for non-existent submission (/admin/view)


jeremythuff:
Currently, from the administrative interface, a request for a non existent submission (i.e. /admin/view/1/1, where submission id 1 does not exist) will result in a the single item view page infinitely attempting to load the non existent submission. 

The console show the following error: "angular.min.js:127 Possibly unhandled rejection: undefined"

The promise awaiting the submissions return should be handled and the user should be forwarded to an appropriate error page.

smutniak:
I'll ask Chris S to review

wwtamu:
@smutniak @cstarcher @jeremythuff this issue can be closed. I have verified that when attempting to go to an admin path in the browser to a non-existing submission it redirects to a submission does not exist view.

![image](https://github.com/TexasDigitalLibrary/Vireo/assets/144841721/54ca4ee0-661b-4eba-aa7f-993b7b027a5a)

The server-side log prints stack trace for the error and should be updated to only log error message when not on DEBUG log level.

```
2023-12-29 10:12:04.927 ERROR 5100 --- [nio-9000-exec-3] o.t.v.controller.SubmissionController    : SubmissionRepoImpl with id 1 not found; nested exception is javax.persistence.EntityNotFoundException: SubmissionRepoImpl with id 1 not found

org.springframework.orm.jpa.JpaObjectRetrievalFailureException: SubmissionRepoImpl with id 1 not found; nested exception is javax.persistence.EntityNotFoundException: SubmissionRepoImpl with id 1 not found
...
```

PR #1443, opened 2020-04-24
V1131: Handle views of non-existent submissions in admin

nwoodward:
Resolves https://github.com/TexasDigitalLibrary/Vireo/issues/1131


nwoodward:
One thing... if this is the correct way to approach handling bad requests, then we'll need to apply it to the other submission controllers.

PR #1457, opened 2020-04-30
[1221] Display customactionvalues properly in excel export

### LINKED ISSUE #1221: Custom Actions do not export correctly in Excel
Custom Actions appear in the Excel Export as
org.tdl.vireo.model.CustomActionValue@20
org.tdl.vireo.model.CustomActionValue@21

Please change so they are similar to Vireo 3 and show which actions have been checked

image

jsavell:
Resolves #1221

### DIFF
--- a/src/main/java/org/tdl/vireo/model/packager/ExcelPackager.java
+++ b/src/main/java/org/tdl/vireo/model/packager/ExcelPackager.java
@@ -8,12 +8,14 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 
 import javax.persistence.Entity;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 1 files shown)


smutniak:
I created 3 custom actions under settings > Application > Custom Actions Checklist and make some of them visible to the student.  I create a couple submissions each under its own user.  I then sign in as admin and export an excel doc.  In the excel doc I see columns A-J with none of them labeled 'Custom Actions' and none of them provide options for a checkbox.  Am I testing this wrong?  Does your test output have more than col A-J?

jsavell:
@smutniak 

Make sure you've added the 'custom actions' field to the list view display under 'customize view', then do the Excel export with Export Columns set to 'Active Filter'.

smutniak:
When I tested this 2 weeks ago I was not seeing the result.  I may just not be testing it correctly.  Since it has been open so long and doesn't appear to have any negative impact I'll merge so it and others can be tested by Stephanie or Christopher.

PR #1459, opened 2020-05-05
[1440] Check map for template string in PackagerUtility to fix sword deposit

### LINKED ISSUE #1440: Deposit Submission to deposit location causes error Hashmap to File cast issue
Dear TDL,

When publishing a submission to a deposit location , the following error occurs at https://github.com/TexasDigitalLibrary/Vireo/blob/1d0e314fa44872463c65eaa653396d217bdf7a5f/src/main/java/org/tdl/vireo/model/depositor/SWORDv1Depositor.java#L122

`java.lang.ClassCastException: java.util.HashMap cannot be cast to java.io.File
	at org.tdl.vireo.model.depositor.SWORDv1Depositor.deposit(SWORDv1Depositor.java:191) ~[classes/:na]
	at org.tdl.vireo.controller.SubmissionController.publish(SubmissionController.java:425) ~[classes/:na]`

When doing the export the file is not a file, but a hashmap. We want to do a DSpaceMets export to our SWORD deposit location. This is unfortunately no possible due to this issue

jsavell:
Resolves #1440

### DIFF
--- a/src/main/java/org/tdl/vireo/utility/PackagerUtility.java
+++ b/src/main/java/org/tdl/vireo/utility/PackagerUtility.java
@@ -22,12 +22,14 @@ public class PackagerUtility {
     @Autowired
     private FormatterUtility formatterUtility;
 
+    private String TEMPLATE_KEY = "template";
+
     public ExportPackage packageExport(Packager<?> packager, Submission submission) throws Exception {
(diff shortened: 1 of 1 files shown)


smutniak:
This appears to be fixed but the dspace instance I am testing against rejects it - likely a mal formed DspaceMETS export and therefore independent of the issue here.

PR #1471, opened 2020-06-08
fix to remove mvn warning about relative path and addition of weaver …

smutniak:
…email dependency for WeaverEmailService

The missing dependency was due to a previous fix but was not evident until running a generated jar in production.


smutniak:
It was open a long time with no commentary or review.  Its only a few lines so I merged it myself.

PR #1492, opened 2020-10-12
4.0.x july2020 staging

smutniak:


cstarcher (approved):
LGTM. Only problem is hardcoded setFrom in SubmissionEmailService.java. I will create an issue to address it.

PR #1504, opened 2020-11-17
[1395][1233][1393][1394][1413] Upgrade Weaver to resolve multiple backend issues

### LINKED ISSUE #1233: A deleted field profile cannot be recreated
Vireo 4 will not allow the re-creation of deleted field profile.   When attempting to Create/Save the new field profile and error message about the field predicate already exists, even if it not in the current workflow.

### LINKED ISSUE #1393: Results vary when ancestor is changed to "Override = No" in ancestor
If a subordinate has "Workflow Step is Overridable" set to No and the ancestor changes the same Workflow Step from yes to "Workflow Step is Overridable" = No, the results will vary.  
Sometimes the subordinate will become Read Only, but not necessarily inherit the current ancestor settings.  Other times it does not become Read Only and retains its settings.

### LINKED ISSUE #1394: Possible bug with Field Profiles and Overridable setting
Changing a Field Profile from “Override = Yes” to “Override = No” on descendant 1 will prevent descendant 2 from changing “Override = Yes” to “Override = No” in same Field Profile.  It will also prevent the Ancestor from changing “Override = Yes” to “Override = No” in same Field Profile.

### LINKED ISSUE #1395: Once a Field Profile has Overridable setting changed to No it cannot be changed back to Yes.
Similar to 732.     When attempting to change the Field profile if you click on Yes then the Update button, the existing Label disappears and the Update button is greyed out.

### LINKED ISSUE #1413: Field profile cannot be moved to another workflow step
I am unable to remove a field profile from one workflow step so that I can place it in another.  The field profile can be deleted, but it cannot be added to the new workflow step, nor can it be recreated in the old workflow step without attempting to sync with parent, which is not working, see #1233.

jsavell:
Resolves #1395 
Resolves #1233 
Resolves #1393 
Resolves #1394 
Resolves #1413


wwelling (approved):

cstarcher:
This resolves each of the five issues listed above. The only problem I'm having is with the page refresh when closing the modal. When I click the "Update" button on the modals, the modal closes, but the screen remains grey and requires a refresh the page. I tested this in both 4.1.0_nov2020_staging and 4.1.0_nov2020-1395-overridable.

cstarcher:
Now that the modal backdrop is being cleared correctly, I can confirm that each of the issues listed above have been resolved with the exception of #1393.

PR #1507, opened 2020-11-21
Actual fix to modal backdrop breaking

wwelling:
Discovered that the modal backdrop still does not close on delete of organization. As the binding has improved during repo updates on broadcast, it still cannot maintain the binding of a selected organization being deleted. 

The actual cause of the modal not closing properly was this ng-if which removed several views when the selected organizations' binding breaks. Changing to ng-show keeps the views on the DOM but hidden, thus allowing bootstrap to close the modal.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "start-dev": "wvr server -b /vireo --uid vireo-webserver"
   },
   "dependencies": {
-    "@wvr/core": "next",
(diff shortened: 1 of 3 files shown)


jsavell (approved):

cstarcher (approved):
This fixes the modal backdrop not clearing on organization delete.

PR #1515, opened 2021-01-19
Migrate from Travis CI to GitHub Actions

kaladay:
Provide a `Build` Workflow for running tests and generating coverage.
This uses a `Build` button to provide the Success/Failure of the Github Actions Workflow.
The word `Build` is used instead of `Coverage` to not conflict with the Coveralls.
The Badge provided by Coveralls will be used for the Coverage because it has the coverage percentage.

Update the badges in the README.md, linking the new `Build` badge to its respective Workflow Actions page.

A simple approach is used and there is room for improvement in the caching approach used.
It is worth considering the `restore-keys` option for the caches.

This utilizes the `testClient` maven build option as well as `dryRun` build option.
The `testClient` build option allows for running the custom `test-ci` build process.
The `dryRun` prevents sending the test results to Coveralls, allowing the Coveralls data to be manually combined.
By manually combining, this alleviates the problems and complexity that sending the Coveralls data in PARALLEL mode introduces.
The program `jq` is used to perform the manual combination.

Currently, the upstream Coveralls Github Action repository does not support using the raw `coveralls.json` file.
Another repository, `MikeEdgar/github-action` (branch `raw_coverage_file`), is providing this behavior.
This commit utilizes this custom repository and branch to more easily transmit the Coveralls data to the Coveralls service.

With these changes, the `travis.yml` can be, and is, removed.

The `package.json` is updated to automatically call `grunt coverage` at the end of the `test-ci` process.
Using `grunt coverage` in this way further simplifies the design of the Github Actions Workflow.

Update the **karma** dependencies.

Use **node** version 10 when running under the Github Actions Workflow.

Github Actions log for the `Build` Workflow may be found here:
https://github.com/TexasDigitalLibrary/Vireo/actions?query=workflow%3ABuild

relates: #1514


wwelling (approved):

pom.xml
```diff
@@ -631,7 +637,7 @@
       <id>test-client</id>
       <activation>
         <property>
-          <name>client</name>
+          <name>testClient</name>
```
wwelling, line 640: Good find.

PR #1534, opened 2021-11-23
List view optimizations

wwelling:
This will break submission list column last event. To fix will likely require a string property be added to submission for last event that is updated on event for the submission. This would require a database migration.

This PR should undergo QA testing due to lack of test coverage. It is possible there will be LazyInitializationExceptions.


kaladay (dismissed):
This shows a significant improvement when using a large dataset and 1000 rows selected.

smutniak (approved):

PR #1535, opened 2021-12-01
List view optimizations band aids

wwelling:


kaladay:
I added a link to the issue but then I realized this PR is into your optimization branch rather than main branch.
I undid the issue link.

PR #1543, opened 2022-03-29
[1525] Remove unique index on Submission, replacing it with a normal index.

### LINKED ISSUE #1525: Preventing multiple degrees from same department
There is a constraint at line 41 in model/Submission.java:
  @Table(uniqueConstraints = @UniqueConstraint(columnNames = { "submitter_id", "organization_id" }))

The constraint above prohibits the migration of theses and dissertations which for people who did multiple works for the same department (e.g. an MS degree in Chemistry followed by a PhD in Chemistry).

If Vireo3 had been used strictly as a tool for processing submissions of theses and dissertations rather than used as a repository then this would not be a problem.
Most sites have not operated this way but have instead treated vireo as an institutional repository.  Migration of data from vireo3 to vireo4 must deal with this historical baggage.

At UTexas this prohibits the migration of hundreds of subsequent theses/dissertations.  (I can get an exact count if needed).  For now, on the utexas-migrate-etd site, I have disabled this constraint.

For what purpose is this constraint needed?  Can this constraint be tightened (submitter_id,organization_id,???) so it doesn't prohibit the scenario above?  Are there other workaround suggestions?

kaladay:
resolves #1525 

Using `submissionDate` is not possible because it can be NULL and is NULL while the submission is In Progress.
A person intending to create two submissions for the same organization might not have at least one submitted.
Furthermore, the granularity of the `submissionDate` is per-day which is too large (it would need to include seconds at the least).

Replace the unique index with a normal index to at least maintain any performance gains of an index while dropping the uniqueness aspect.

The UI also needs changes to allow for creating a new submission.
The pre-existing behavior always tries to continue editing if an existing submission has the same organization for a new submission.
This makes it impossible to create a new submission for the same organization as an existing submission.
The if condition logic is changed such that only when the existing submissions organization no longer allows submissions shall it perform a "continue submission" rather than creating a new submission.

This introduces a new situation that did not exist before that needs additional consideration that is (partially) outside the scope of this change.
When there are two or more existing submissions for the same organization and that organization no longer allows submissions then there is way to select which of the existing submissions to continue.
This requires design change considerations.
Rather than make these decisions within the scope of this issue, I have opted to have this problem addressed in a separate issue.
The current implementation preserves the existing nuances where "continue" should still happen as it did before, so long as there is a single submission for some organization or that organization allows new submission.

The tests have been updating accordingly because the pre-existing tests are expecting uniqueness.
The `findBySubmitterAndOrganization()` method is only used by tests and has been update to `findAllBySubmitterAndOrganization()`.

### DIFF
--- a/src/main/java/org/tdl/vireo/model/Submission.java
+++ b/src/main/java/org/tdl/vireo/model/Submission.java
@@ -26,22 +26,24 @@ import javax.persistence.Temporal;
 import javax.persistence.TemporalType;
 import javax.persistence.UniqueConstraint;
 
-import com.fasterxml.jackson.annotation.JsonIgnore;
-import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
-import com.fasterxml.jackson.annotation.JsonView;
-
 import org.hibernate.annotations.Fetch;
 import org.hibernate.annotations.FetchMode;
 import org.tdl.vireo.model.response.Views;
 import org.tdl.vireo.model.validation.SubmissionValidator;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonView;
 
 import edu.tamu.weaver.validation.model.ValidatingBaseEntity;
 
 @Entity
 @JsonIgnoreProperties(value = { "organization" }, allowGetters = true)
 @Table(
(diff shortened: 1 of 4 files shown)


wwelling (approved):
I suspected our quick impromptu planning on this would have oversights. My primary concern was of performance and could not think of how it could be a breaking change without the constraint. Hopefully, we did not code with the expectation of this constraint. I do think this is an acceptable solution for the issue.

wwelling:
The JPA named method only used by tests probably should just be removed as dead code. But fine with it and updated tests.

cstarcher (approved):
Works well.

PR #1544, opened 2022-03-29
Perform submission delete within session transaction

### LINKED ISSUE #1540: Error when deleting submission in Submission History
When deleting a submission in Submission History, the submission is removed, but I receive the following error:

Screen Shot 2022-03-23 at 2 39 51 PM

cstarcher:
resolves #1540

### DIFF
--- a/src/main/java/org/tdl/vireo/controller/SubmissionController.java
+++ b/src/main/java/org/tdl/vireo/controller/SubmissionController.java
@@ -215,6 +215,7 @@ public class SubmissionController {
     return new ApiResponse(SUCCESS, submission);
   }
 
+  @Transactional
   @RequestMapping("/delete/{submissionId}")
   @PreAuthorize("hasRole('STUDENT')")
   public ApiResponse deleteSubmission(@WeaverUser User user, @PathVariable Long submissionId) {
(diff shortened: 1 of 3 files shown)


kaladay:
There is also the possible approach of making changes to the UI to, when these properties are needed manually load the additional data. This would be a lot more changes.

I am going to test the before/after to see if this causes any performance problems given that this partially reverts the EAGER to LAZY performance fixes.
Hopefully there are no significant performance losses because making the changes here is much easier than the alternate changes to the UI.

kaladay:
The Before and After screenshot showing the performance difference as of [fec64c8](https://github.com/TexasDigitalLibrary/Vireo/pull/1544/commits/fec64c80579b61fcc16d3cd5943dfbc5a33cc99c).

Screenshot_2022-03-29_16-01-55

cstarcher:
Clearly, this is not the ideal solution in light of previous performance issues concerning submission lists. I am investigating alternative solutions to avoid performance degradation.

wwelling:
@cstarcher have you tried adding @Transactional annotation to the controller endpoint?

wwelling:
Try adding this, https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/transaction/annotation/Transactional.html

here
https://github.com/TexasDigitalLibrary/Vireo/blob/main/src/main/java/org/tdl/vireo/controller/SubmissionController.java#L216

kaladay (approved):
Looks great.
~400ms wait with a ~3ms receive.

I changed the list size to ~1000 and had a ~500ms wait with a ~1.5s receive.

smutniak (approved):

wwelling (approved):

Issue #1522, opened 2021-01-27
Who Has Access in Settings is not populated


smutniak:
Settings>Application> Who Has Access > Add Member    button produces error

cstarcher:
Further explanation by user:

I should be able to see who has access in the Settings area (Application Settings à Who Has Access).  I’ve attached a couple of screenshots, which clearly shows nothing in the list.
 
I believe – based on Vireo 3 – I should be able to see myself as anyone else who is an Admin or Manager.  And when I click on the “Add Member” button, I should also be able to see anyone with an account and set admin privileges.
 
As Frank mentioned, I can see a list of all users in the Manage Users drop-down and be able to promote/demote users to admin/manager roles from there.  But the Who Has Access view should show only the people who have the roles already assigned (without having to go through the full list).

![Screenshot 2020-12-09 14 17 45](https://user-images.githubusercontent.com/5767401/119224989-cb2dcf00-bac6-11eb-8218-aa22efc4554e.png)
![Screenshot 2020-12-09 14 18 45](https://user-images.githubusercontent.com/5767401/119224990-cbc66580-bac6-11eb-8c59-95001e7e7b39.png)

PR #1546, opened 2022-03-31
Improve angularjs tests for user.

kaladay:
The mock.user mock is not mocking in the correct way.
When a `new User()` command is called mocks fail.

Instead, have mock.user return a function, causing User to act as a constructor and allowing for `new User()`.
The original mockedUser must then be returned by the mocked constructor.

This exposed a bug in one of the tests, which was rewritten.

With this change, the `$injector.get()` no longer has problems and can be used.
It should be used and now is used for the repo tests.

These problems were discovered (and fixed in) while working on #1522.


smutniak:
Looks good.  I presume the conflict came from other work you did on another ticket, correct?

smutniak (approved):

kaladay:
> Looks good. I presume the conflict came from other work you did on another ticket, correct?

The referenced issue exposed this problem and I discovered the mistakes.

These are from previous work I did quite a while ago and I apparently didn't do as good of a job as I could have.

cstarcher (approved):

PR #1547, opened 2022-03-31
[1522] Who Has Access in Settings is not populated.

kaladay:
There are several problems observed and fixed within the scope of this issue.

Summary:

1. "Assignable" users in backend is not including `ROLE_REVIEWER`.
2. Only include not assignable users in *Add Member* (referred to as _unassignable_ users).
3. The userRepoController.js is being double-purposed and there needs to be a separate controller for "Who Has Access".
4. Improve efficiency of fetched data.
5. Remove redundant UserRepo.listen() call.
6. Remove redundant modal views associated with the *Add Member* modal.
7. Perform syntax clean ups in the related code.
8. Note the problems with listener and changes.

The backend is supposed to include `ROLE_REVIEWER` when determining assignable users.
This is obvious because the *Add Member* modal adds users by assigning them `ROLE_REVIEWER`.
Because the `ROLE_REVIEWER` is not being included, the added users simply disappear.

The *Add Member* modal should only include those users that are not currently assignable.
To keep the terminology consistent, I have opted to name these _unassignable_.
A new endpoint is created and used that only includes `ROLE_STUDENT` and `ROLE_ANONYMOUS`.
This should be more performant that loading every user in the backend and then filtering them in the *Add Member* modal for loops.

The "Who Has Access" stopped working because of the following commit and line number:
- https://github.com/TexasDigitalLibrary/Vireo/commit/dc6973b61c39276d70949e2b0499d3009bfa8129#diff-2a52b015369321d4b3a7b0bd36dc45612a66e24c0df031e279ad35b4c1e04516L7

This is also relevant:
- https://github.com/TexasDigitalLibrary/Vireo/commit/989098e4c35393b38946899f187c2bc3b92f82b8

The loss of `$scope.users` is the result of the `userRepoController.js` performing double duty.
This can be avoided by adding a new controller, now called `whoHasAccessController.js`.

This is an opportunity to make the code a little bit more efficient.
Rather than fetching all users, only fetch the assignable users.
When the *Add Member* modal is opened, then and only then fetch only all of the unassignable users.

The UserRepo.listen() is defined twice.
The first case only closes an open modal.
The second case does other things.
While I question what those other things are, I am considering such questions out of scope and I am ignoring them.
Remove the first case and have the second case also perform the close modal.

There are multiple *Add Member* modal views:
- app/views/modals/addMemberModal.html
- app/views/modals/whoHasAccessModal.html

These are almost identicaly, ignoring syntax and styling differences.

Because the `addMemberModal.html` has a more accurate name, I chose to delete `app/views/modals/whoHasAccessModal.html`.

As for the " Note the problems with listener and changes." topic...
It seems this problem affects the users in general.
If I go to the "Manage Users", the updates do not broadcast or get received either.
Therefore, I conclude that the problems with the listeners and updating on changes is a separate issue.
A new issue should and will be created for this if one does not already exist.

This also includes fixes for tests.
These fixes were also applied across the rest of the tests in #1546.


cstarcher (changes requested):
Two issues:

The modal is not updating when adding a user. I would expect that the selected user would be removed from the list after clicking the add button. The user is, however, added to the list.

The list does not update when assigning roles for STUDENT and ANONYMOUS. I would expect the user to be removed after selecting one of these roles. It only updates after I refresh the page.

cstarcher (approved):

cstarcher (approved):
Since the broadcast issue is more widespread, I am fine with addressing it in a different issue and adding it to the sprint.

PR #1551, opened 2022-03-31
changed preauthorize /embargo/all path to anonymous

### LINKED ISSUE #1365: "Review Application", 401 error when not logged in
The "Review Application" path allows for one to view the page with a valid hash without otherwise having access to the system.
The problem is that when a user is not actually logged in, the `ERROR: (401) You are not authorized for request ` is improperly presented.
The error should not be presented in this case or the access credentials should not be checked at this path.

cstarcher:
Resolves #1365 

This fixes the error. Are there any security concerns about this fix?

### DIFF
--- a/src/main/java/org/tdl/vireo/controller/EmbargoController.java
+++ b/src/main/java/org/tdl/vireo/controller/EmbargoController.java
@@ -34,7 +34,7 @@ public class EmbargoController {
     private EmbargoRepo embargoRepo;
 
     @GetMapping("/all")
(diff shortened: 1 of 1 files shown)


kaladay (changes requested):
I checked with the team here and this should be fine to be publically available given the tokenied URL path.

However, if we are making this public, I suggest entirely removing the `@PreAuthorize` line rather than explicitly setting it to `ANONYMOUS`.

kaladay (approved):

smutniak (approved):

PR #1553, opened 2022-04-04
[1274] Redesign email handling to avoid race condition between Weaver and Spring.

### LINKED ISSUE #1274: Configuration e-mail from address might not be used
I noticed that the **from** address in e-mails sent using Vireos _comment_ or _batch/comment_ interface are not using the `app.email.from` values found within `src/main/resources/application.yaml`.

I am using a Linux host and it is provide the local systems _username@hostname_ as the from address instead of the one specified by the configuration.

I am uncertain on whether or not this is an issue because I am testing using `mvn clean spring-boot:run` or if that configuration does not apply.

kaladay:
resolves #1274

This is branched off of #1545, please review and ideally approve that PR prior to this one.

There is a race condition during the loading data during sprint boot.

The `EmailSender` Bean may be loaded either before or after the Spring `@Value` annotations are loaded.
If `EmailSender` is loader after the `@Value` annotations are loaded, then the e-mails works as expected.
If the opposite happens, then no data is loaded and the Weaver default (which uses a mailinator address) is incorrectly used.

To avoid this, this commit does the following:
1) Do not use any `@Value` annotations (or any Spring configuration loaders) in the `AppEmailConfig` (where the Bean is loaded).
2) Implement a custom AppConfig loader, called `VireoEmailConfig`, to handle loading of the data.
3) Have the `SubmissionEmailService` Autowire the `VireoEmailSender` and use that to send the e-mails.
4) Load the settings for each e-mail (not ideal, future work is recommended to improve this).

The encoding format shouldn't be hard coded when everything else is not.
Furthermore, it seems that the Weaver design doesn't appear to set the encoding properly (Code shows `UTF-8` and actual e-mail headers show `us-ascii`).

The `VireoEmailConfig` now needs to be available for testing.
The Mockito `verify()` static method does a bad job of detecting if something actually is called.
Even though the final `sendEmail()` method that has 5 arguments gets called (thanks to adding the `doCallRealMethod()`), the `verify()` fails to recognize this.
The tests have to be changed to handle the specific function call even if it is just a pass through method.

### DIFF
--- a/src/main/java/org/tdl/vireo/auth/controller/AuthController.java
+++ b/src/main/java/org/tdl/vireo/auth/controller/AuthController.java
@@ -16,6 +16,7 @@ import java.util.concurrent.TimeUnit;
 import javax.crypto.BadPaddingException;
 import javax.crypto.IllegalBlockSizeException;
 import javax.crypto.NoSuchPaddingException;
+import javax.mail.MessagingException;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -30,6 +31,7 @@ import org.tdl.vireo.model.EmailTemplate;
 import org.tdl.vireo.model.User;
 import org.tdl.vireo.model.repo.EmailTemplateRepo;
 import org.tdl.vireo.model.repo.UserRepo;
+import org.tdl.vireo.service.VireoEmailSender;
 import org.tdl.vireo.utility.TemplateUtility;
 
 import edu.tamu.weaver.auth.controller.WeaverAuthController;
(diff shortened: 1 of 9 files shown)


cstarcher (approved):
I tested batch emails, email workflow rules, and generate advisor email. All seem to be assigning email config addresses.

I did run across a separate issue with add comment emails not being sent correctly when using placeholders, but it is also behaving the same way in 4.1.2, so I don't think it's related to this fix. However, I will mention this PR in the new issue for added context.

PR #1560, opened 2022-04-12
[1555] Email placeholders in Add Comment and Add File modals not working.

### LINKED ISSUE #1555: Email placeholders in Add Comment and Add File modals not working
Using placeholders, such as Committee Chair, in the Add Comment and Add File Modals is not working.

Screen Shot 2022-04-05 at 1 30 11 PM

Relates to Issue #1431 
Adding relation to Issue #1274 for context regarding recent changes to email behavior

kaladay:
resolves #1555 

This is primarily a regression caused by the Lazy loading performance improvements.
Because the data is lazily loaded, it is not available and no e-mails are obtained to be sent.
Utilize a graph to force Eager loading in the explicit case of e-mails.
This required:
- Relocating controller logic code into the relevant service model and performing the database load there.
- Utilizing the attribute path entity graphs, see links below for details.
- have the controller instead pass the submission id onto the e-mail service methods.

Add the BCC to the e-mail debug log as well.

see: https://www.baeldung.com/spring-data-jpa-named-entity-graphs
see: https://www.baeldung.com/spring-open-session-in-view

### DIFF
--- a/src/main/java/org/tdl/vireo/controller/SubmissionController.java
+++ b/src/main/java/org/tdl/vireo/controller/SubmissionController.java
@@ -27,11 +27,6 @@ import java.util.zip.ZipOutputStream;
 import javax.servlet.ServletOutputStream;
 import javax.servlet.http.HttpServletResponse;
 
-import com.fasterxml.jackson.annotation.JsonView;
-import com.fasterxml.jackson.core.JsonProcessingException;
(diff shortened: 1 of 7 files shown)


cstarcher (approved):

smutniak (approved):
looks good.  As infrequently as it is used I don't expect the flicker will bother anybody.

PR #1561, opened 2022-04-12
[1549] UI: Too many users, needs pagination to limit as well as a paginated table for "Who has Access" and "Add Member".

### LINKED ISSUE #1549: UI: Too many users, needs pagination to limit as well as a paginated table for "Who has Access" and "Add Member".
With the "Who has Access" now working again thanks to PR #1547, the `Who has Acces` and `Add Member`.

The `Manage Users` page already has pagination so it would make a good example for implementing this.

For large user bases I consider this a bug as it almost crashes my browser when loading ~2000 users with `ROLE_STUDENT`.
Because there is no existing pagination for these particular areas, I consider this a new feature added that will solve a bug.

kaladay:
resolves #1549 

Change existing endpoints and add new endpoints.
- /assignable
- /assignable/total
- /unassignable
- /unassignable/total

These new endpoints now support pagination.

Rather than changing the UI internal design, I opted to just create a new endpoint for getting the totals.
These are needed for proper pagination support.

There is a work-around (see code comments) that allows for Pageable to have a default value of 0.
This means load all results without pagination.
In such a case, filtering is still supported.

In order to achive the filter, a synthetic DB column "name" is generated and is including at the DB level.
This is achieved by utilizing `@Formula` annotation.
This annotationr equired that I explicitly set the name annotation property for the firstName and lastName User properties.

Several JPA functions are added to accomplish all of the necessary tasks.

The UI loads the totals first and if the total is non-zero, then loads the actual data.

The existing filter from Add Member is removed in favor of the ng-table filter.

There is a lot more cleanup that could be done but for brevity and time-limits these potential improvements are ignored.

Problems that are not resolved:
- There is a flicker on filter changes or pagination that is annoying.
- Changes to either form still require full page refreshes.

### DIFF
--- a/src/main/java/org/tdl/vireo/controller/UserController.java
+++ b/src/main/java/org/tdl/vireo/controller/UserController.java
@@ -6,19 +6,25 @@ import static edu.tamu.weaver.validation.model.BusinessValidationType.UPDATE;
 import static org.springframework.beans.BeanUtils.copyProperties;
 import static org.springframework.web.bind.annotation.RequestMethod.POST;
 
-import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 
+import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.data.domain.Pageable;
+import org.springframework.data.domain.Sort.Direction;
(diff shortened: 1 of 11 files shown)


cstarcher (approved):
This solves Issue #1549. I will create an issue for each of the remaining problems identified by @kaladay.

smutniak (approved):
looks good.  As infrequently is it is used I don't expect the flicker will be a problem

PR #1564, opened 2022-04-14
4.1.2 spring2022 staging

smutniak:
I'll create the PR now for when the large data test completes.


smutniak:
The biggest data I could test on was TTU's which is 2/7 the size of UTexas and the second biggest data set.  It performed ok. 

My test was to pull up a list and change the number of items per page and observe how long it took.  I didn't measure quantitatively just if it was tolerable or noticable.  It maybe took a second or 2 with the TTU data.   With baylor data (about a third of ttu data it seemed about a half second). Also note that this was on a 2019 i7 mac with 16gb in a VM.  The AWS instances can be given more resources if need be.

I'll still test utexas data at some point but their deploy is a ways off and everything looks great for the sites we are deploying now and in the near future.

cstarcher (approved):

kaladay (approved):

PR #1604, opened 2023-03-13
4.1.3 excelexport 1581

smutniak:
Adds data to excel export columns SubmissionStatus per the ticket.  AssignedTo data was also added.


smutniak:
Need to clean out a few things I noticed late.   will

Issue #1588, opened 2023-01-30
DSpaceMETS export cannot handle multiple files with similar names


smutniak:
DSpaceMETS export, because of the getFileName response of FieldValue, sees some file names as duplicates because the prefixes of files are truncated.  This is particularly a problem where multiple archive files exist and the unique prefixes of the files are stripped and they then are reported to have the same name.

PR #1606, opened 2023-03-14
4.1.3 d space mets 1588

smutniak:
This fixes DSpaceMETS export problems - it builds out missing functionality and uses the full names of certain files
so they do not appear as duplicates to the zip utility.  
This also fixes ProQuest exports lack of data in some sections.

DSPACEMETS
For an DSpaceMETS export of keywords (needed by some sites) the keywords
field predicate should be known as 'dc.subject.other' 
 
PROQUEST
For a complete set of data in a ProQuest export there needs to be
  values in:
        degree.degree_code
  values in:
        field_value.definition
        field_value.identifier
  where field_predicate.value = 'thesis.degree.name'

These values can be set manually for testing.

For sites in production, if the values are not there then they can be set
using FixProquestExport.rb in the migration repo
(which for now is not publicly available but I can send)


smutniak:
resolves #1588

wwelling (approved):

PR #1634, opened 2023-03-24
Initial "simple models" design.

kaladay:
Attempts have been made to use EntityGraphs and Projections. Strange problems have been encountered.
Rather than trying to investigate those, I opted for this alternative approach.

This utilizes immutable entities that act similar to an SQL view. These immutable entities have expensive columsn removed or otherwase made to not load (even lazy load). Consider this the concept of a non-existent third FetchType called "NEVER" (compared to FetchType.EAGER and FetchType.LAZY).

There have been problems observed with Spring or JPA not being able to handle certain column stuctures with the immutabille design approach. These have been made Transient to prevent the problems. The downside is that these now Transient columns must be fetched. This performs those extra manual fetches in circumstances where they will be needed. Such fetches are done on the backend to limit the amount of changes to the UI.

In many cases, these simple models are being cast to the normal model before sending them to the frontend/UI. This required some additional tweaks.
The UI is still modified to handle the simple submission model in case that comes down rather than the normal submission model.

The simple models, being immutable, somehow interefered with the normal models. The table name has been explicitly added in these cases. The table names are named after how they have been observed to be created.

To keep these unusual changes isolated, I created directories called "simple". Inside these are the simple models and the respective simple repositories. The simple repositories have been created on an as-required basis. These simple repositories are made as simple as possible and only contain an interface file.

The last event is now handled during processing of the simple submission model handling. This avoids an extra db check and avoids extra logic beyond the initial handling.

This does not address the big custom crafted select all submissions. Additional work is needed to potentially take advantage of the changes here for that big query.

NULL checks and NULL values are added to handle cases where the data is not loaded (and is not FetchType.LAZY).

NULL checks are added to the UI to handle cases where the data is not available now that everything is not loaded.

This has only been tested against postgresql so far. The other DBs need to be tested.

### DIFF
--- a/src/main/java/org/tdl/vireo/controller/SubmissionController.java
+++ b/src/main/java/org/tdl/vireo/controller/SubmissionController.java
@@ -240,6 +240,7 @@ public class SubmissionController {
         SimpleSubmission submission = simpleSubmissionRepo.findByAdvisorAccessHash(advisorAccessHash);
         submission.setFieldValues(simpleFieldValueRepo.findAllBySubmissionId(submission.getId()));
         submission.setActionLogs(actionLogRepo.findAll(submission.getId()));
+        submission.setSubmissionWorkflowSteps(submissionWorkflowStepRepo.findBySubmissionId(submission.getId()));
 
         return new ApiResponse(SUCCESS, SimpleSubmission.toSubmission(submission));
     }
--- a/src/main/java/org/tdl/vireo/controller/SubmissionController.java
+++ b/src/main/java/org/tdl/vireo/controller/SubmissionController.java
@@ -24,6 +24,7 @@ import java.nio.file.Path;
 import java.text.DateFormat;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
+import java.util.ArrayList;
 import java.util.Calendar;
 import java.util.HashMap;
 import java.util.List;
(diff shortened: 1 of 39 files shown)


src/main/java/org/tdl/vireo/controller/SubmissionController.java
```diff
@@ -101,142 +106,178 @@
...
 
-    ApiResponse response = new ApiResponse(SUCCESS);
-    if (submissionToDelete.getSubmitter().getEmail().equals(user.getEmail())
-        || user.getRole().ordinal() <= Role.ROLE_MANAGER.ordinal()) {
-      submissionRepo.delete(submissionToDelete);
-    } else {
-      response = new ApiResponse(ERROR, "Insufficient permisions to delete this submission.");
+        submission.setFieldValues(simpleFieldValueRepo.findAllBySubmissionId(submission.getId()));
+        submission.setActionLogs(actionLogRepo.findAll(submission.getId()));
+        submission.setSubmissionWorkflowSteps(submissionWorkflowStepRepo.findBySubmissionId(submission.getId()));
```
wwelling, line 233: Submission workflow steps are ordered per submission using `OrderColumn` annotation on the model. The `findBySubmissionId` method will have to order the same.

PR #1637, opened 2023-03-27
Issue 1596: Performance improvements using `EntityGraph` and `JsonView`

wwelling:


kaladay (approved):
This is good to go.
I will be performing additional runtime testing and I will follow up with additional PRs for any regressions found as needed.

PR #1654, opened 2023-03-29
Restrict unused endpoint.

kaladay:
We should probably decide on whether or not to keep this endpoint.

This endpoint could be changed to be paginated et al. and be used for the list if more advanced uses of the forms are provided. Or this could be removed entirely.

This takes the approach of restring this to the ADMIN use to avoid potential abuses until such a time a decision can be made.


wwelling (approved):
Thanks. Why provide admin to that endpoint again? That enables use in production. It should be configurable to avoid crashing either the database or the service or any open browsers for that matter. Or we measure how many admins can be using it simultaneous without affecting student use?

PR #1677, opened 2023-04-04
Issue 1662: Improve accessibility regarding footer links, providing more details.

### LINKED ISSUE #1662: Improve link text names/descriptions in footer
Relevant Components:

- [x] Home
- [ ] File Upload

Description
The links in the footer have inaccessible names. The link destination should be clear from the accessible name. For example, “send an email to the webmaster” is more descriptive than “webmaster.”

Current Code
```<a href=…>Webmaster</a>```

Solution
Make sure the link destination is explicit, if not in the displayed label, then in the accessibility tree.
Recommended Code
```<a href=… aria-label=”Send an email to the webmaster”>Webmaster</a>```

kaladay:
Resolves #1662 

The links are dynamically generated and have three possibilities:
1) A link to an e-mail.
2) A link to a URL.
3) A self-link.

The message must be accurate to the purpose so programmatically generate the text.

In the third case, the description cannot be provided because there is no link anywhere. I believe this may have been a design decision to preserve the link's presence. In this specific case, the link has no purpose and the best way to communicate this is to tell the reader to ignore this element.

Ideally, one time binding would be used here as these values should not regularly change. The one time binding does not appear to be supported for the `ng-attr` functionality.

Logic inside the html is to be avoided and so properties like `accessibility_hidden` are created to trigger the inverse case when `accessibility_label` is empty and `aria-hidden` is to be used.

### DIFF
--- a/src/main/resources/templates/index.html
+++ b/src/main/resources/templates/index.html
@@ -153,16 +153,16 @@
           <span app-version></span>
         </li>
         <li>
-          <a href="{{webmaster}}">Webmaster</a>
+          <a href="{{webmaster}}" ng-attr-aria-label="{{webmaster_label}}" ng-attr-aria-hidden="{{webmaster_hidden}}">Webmaster</a>
         </li>
         <li>
-          <a href="{{legal}}">Legal</a>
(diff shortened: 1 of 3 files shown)


cstarcher (approved):
Is dynamic and works for both email and domain name entries.

PR #1680, opened 2023-04-05
Issue 1608: Add new end points for action log and re-design code depending on new end points.

### LINKED ISSUE #1608: Faculty Review URL broken/confusing table when there are private action logs.
The Faculty Review page for a submission explicitly only shows public action logs.

The private comments are still taking up space causing the table.
The table is counting them but then not displaying them when private.
This can cause weird structures such as a table of only 3 rows then a table of 9 rows when 10 are consistently always expected.

kaladay:
Resolves #1608

Add two new end points:
1) `/submission/get-one/{submissionId}/action-logs`
2) `/submission/advisor-review/{advisorAccessHash}/action-logs`

These are placed within the submission controller as there is no action log controller and the path structure is within the submission path. Two end points are created because the fetching by hash is not the only situation where the action logs are fetched for some submission. The student submission view also has the same exact problem as with the advisor page. The student submission view uses the submission ID as requires a logged in account whereas the advisor end point uses a hash and does not require a logged in user.

Two new matching repo methods are created to handle this behavior. The by hash fetching is more advanced in that it needs to join submission table to properly match the hash. I made several attempts to use the JPQL for `@Query` and I constantly ran into problems starting up. I decided not to pursue this problem further for time-constraint reasons and move on. I used native queries as a work-around, which is undesirable.

The UI action log table directive is updated to optionally use a callback rather than data straight off of the submission. The pagination logic is passed directly through to the callback and the response is handled appropriately. If the callback is not provided then the default behavior of processing the action logs off the submission is preserved.

Now that action logs are not always being loaded, in certain cases, additional undefined/NULL checks are required in the UI.

Unit tests are partially written.
There are no Submission Controller unit tests and so a completely new file is created and only the functions added here have tests written for them. The repo methods are using `@Query` and so there is nothing to test for those methods. The UI unit tests are written but are incomplete.
I see no existing pattern that has similar code fully working in a unit test and so a bare minimal test is written.

Implement a new JsonView called `SubmissionIndividualActionLogs`. This allows for having a `SubmissionIndividual` without the action logs being lazily loaded. The original behavior is preserved where needed by calling the newly minted `SubmissionIndividualActionLogs` instead of `SubmissionIndividual`.

### DIFF
--- a/src/test/java/org/tdl/vireo/controller/SubmissionControllerTest.java
+++ b/src/test/java/org/tdl/vireo/controller/SubmissionControllerTest.java
@@ -1,0 +1,96 @@
+package org.tdl.vireo.controller;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.when;
+
+import edu.tamu.weaver.response.ApiResponse;
+import edu.tamu.weaver.response.ApiStatus;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Calendar;
+import java.util.GregorianCalendar;
+import java.util.List;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.InjectMocks;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+import org.springframework.data.domain.Page;
+import org.springframework.data.domain.PageImpl;
+import org.springframework.data.domain.PageRequest;
+import org.springframework.data.domain.Pageable;
+import org.springframework.test.context.ActiveProfiles;
+import org.springframework.test.context.junit.jupiter.SpringExtension;
+import org.tdl.vireo.model.ActionLog;
+import org.tdl.vireo.model.Role;
+import org.tdl.vireo.model.SubmissionState;
+import org.tdl.vireo.model.SubmissionStatus;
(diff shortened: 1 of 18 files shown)


cstarcher (approved):

src/test/java/org/tdl/vireo/controller/SubmissionControllerTest.java
```diff
@@ -0,0 +1,96 @@
...
+import org.tdl.vireo.model.SubmissionStatus;
+import org.tdl.vireo.model.User;
+import org.tdl.vireo.model.repo.ActionLogRepo;
+
+@ActiveProfiles("test")
+@ExtendWith(SpringExtension.class)
+@ExtendWith(MockitoExtension.class)
+public class SubmissionControllerTest {
+
+    private static final String TEST_USER_1_EMAIL = "User 1 email";
```
wwelling, line 38: Move to a JSON file as per best practice.

wwelling:
Please describe the existing endpoints for action logs and why not refactor it with appropriate query parameters? The reviewer should be an administrator, chair, co-chair, and unfortunately whomever the administrator shares a link with. The logs presented should be based on the role which is in the JWT. That can be retrieved in any controller endpoint via a weaver annotation.

src/test/java/org/tdl/vireo/controller/SubmissionControllerTest.java
kaladay: I'm going with the existing design practices by Vireo.
I'm not out to change the practices.
If there is a change then we need to do it across the entire Vireo project.
I suggest creating an issue for changing the test code practices.

PR #1681, opened 2023-04-05
Issue 1632: Allow file deletion to work even when the file on the disk is not found.

### LINKED ISSUE #1632: Cannot delete file via web interface if file is missing on the disk.
During testing, I uploaded a file.
Then I restarted the the system while preserving the DB tables and respective data but I cleared the local file cache.
This caused the situation where the file is referenced but cannot be found.

Example of broken file showing NaN undefined

I figured I could just delete it.
It won't let me delete the file because the local file is not found.

```
ERROR: (500) vireo/target/classes/private/587811713/1679688528996-DANIELS-PRIMARY-2023.pdf 
```

This is a request to not error out on delete if the file is missing from the disk.
Just treat it as a warning and remove the reference from the DB.

This is observed on the Student Submission View but may very well be affecting all of the file removals on the system.

kaladay:
Resolves #1632 

If the file no longer exists during the delete process then do not error out. Instead, print a warning to the console.

There error actually first happens on `assetService.getAssetFileInfo()` which does not check to see if the file exists. Modifying that code appears to have larger affects and is therefore not modified. New functions are instead provided to allow for deletion without erroring on non-existent file.

Report that the file is not found in the action log where the file size would otherwise be.

Screenshot showing the action log message

### DIFF
--- a/src/main/java/org/tdl/vireo/controller/SubmissionController.java
+++ b/src/main/java/org/tdl/vireo/controller/SubmissionController.java
@@ -990,9 +990,18 @@ public class SubmissionController {
             apiResponse = new ApiResponse(ERROR, "You are not allowed to delete license files!");
         } else {
             if (user.getRole().equals(Role.ROLE_ADMIN) || user.getRole().equals(Role.ROLE_MANAGER) || uri.contains(String.valueOf(hash))) {
-                JsonNode fileInfo = assetService.getAssetFileInfo(uri, submission);
+                String fileName = "";
+                String fileSize = "file not found";
+                if (assetService.assetFileExists(uri)) {
+                    JsonNode fileInfo = assetService.getAssetFileInfo(uri, submission);
+                    fileName = fileInfo.get("name").asText();
+                    fileSize = fileInfo.get("readableSize").asText();
+                } else {
+                    fileName = assetService.getAssetFileName(uri);
+                }
+
                 assetService.delete(uri);
-                actionLogRepo.createPublicLog(submission, user, documentType.substring(9).toUpperCase() + " file " + fileInfo.get("name").asText() + " (" + fileInfo.get("readableSize").asText() + ") removed");
+                actionLogRepo.createPublicLog(submission, user, documentType.substring(9).toUpperCase() + " file " + fileName + " (" + fileSize + ") removed");
             } else {
                 apiResponse = new ApiResponse(ERROR, "This is not your file to delete!");
             }
--- a/src/main/java/org/tdl/vireo/service/AssetService.java
+++ b/src/main/java/org/tdl/vireo/service/AssetService.java
@@ -1,5 +1,7 @@
 package org.tdl.vireo.service;
 
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import java.awt.image.BufferedImage;
 import java.io.ByteArrayInputStream;
 import java.io.File;
@@ -25,11 +27,11 @@ import java.util.HashMap;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
-
 import javax.imageio.ImageIO;
-
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.IOUtils;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.core.io.Resource;
 import org.springframework.core.io.support.ResourcePatternResolver;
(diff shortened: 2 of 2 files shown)


jcreel (approved):
Handles deletion of asset entity from submission, regardless of assets on disk, with robust logging.

cstarcher (approved):

src/main/java/org/tdl/vireo/controller/SubmissionController.java
```diff
@@ -990,9 +990,18 @@ public ApiResponse removeFile(@WeaverUser User user, @PathVariable Long submissi
             apiResponse = new ApiResponse(ERROR, "You are not allowed to delete license files!");
         } else {
             if (user.getRole().equals(Role.ROLE_ADMIN) || user.getRole().equals(Role.ROLE_MANAGER) || uri.contains(String.valueOf(hash))) {
```
wwelling, line 992: Why does manager or admin get to delete a file uploaded? That runs the risk of a student uploaded a file from wherever (laptop, cellphone, lab, home personal computer) and since uploading lost all work for unknown reasons. This scenario should be addressed and a communication is required before any file deletion!

wwelling: If it is not your account, you cannot change any interaction with Vireo. Otherwise, the application is suspicious of nefarious activity.

src/main/java/org/tdl/vireo/controller/SubmissionController.java
```diff
@@ -990,9 +990,18 @@ public ApiResponse removeFile(@WeaverUser User user, @PathVariable Long submissi
...
-                JsonNode fileInfo = assetService.getAssetFileInfo(uri, submission);
+                String fileName = "";
+                String fileSize = "file not found";
```
wwelling, line 994: Assuming file not found is not necessary as the file should only be deleted and visible before submission complete. Administrators should be prevented from viewing the submission until it is complete unless assistance requested from the advisory office.

src/main/java/org/tdl/vireo/controller/SubmissionController.java
```diff
@@ -990,9 +990,18 @@ public ApiResponse removeFile(@WeaverUser User user, @PathVariable Long submissi
...
+                if (assetService.assetFileExists(uri)) {
```
wwelling, line 995: Reasonable method but why condition. Shouldn't application know the file exists already?

src/main/java/org/tdl/vireo/controller/SubmissionController.java
```diff
@@ -990,9 +990,18 @@ public ApiResponse removeFile(@WeaverUser User user, @PathVariable Long submissi
...
+                    JsonNode fileInfo = assetService.getAssetFileInfo(uri, submission);
+                    fileName = fileInfo.get("name").asText();
+                    fileSize = fileInfo.get("readableSize").asText();
+                } else {
+                    fileName = assetService.getAssetFileName(uri);
```
wwelling, line 1000: Why present a filename to the front end at all? Who deleted the file if the user was unaware of it?

src/main/java/org/tdl/vireo/controller/SubmissionController.java
```diff
@@ -990,9 +990,18 @@ public ApiResponse removeFile(@WeaverUser User user, @PathVariable Long submissi
...
+                }
+
                 assetService.delete(uri);
-                actionLogRepo.createPublicLog(submission, user, documentType.substring(9).toUpperCase() + " file " + fileInfo.get("name").asText() + " (" + fileInfo.get("readableSize").asText() + ") removed");
+                actionLogRepo.createPublicLog(submission, user, documentType.substring(9).toUpperCase() + " file " + fileName + " (" + fileSize + ") removed");
```
wwelling, line 1004: Good addition to the log.

src/main/java/org/tdl/vireo/service/AssetService.java
```diff
@@ -87,7 +88,12 @@ public String find(String relativeFolderPath, String name) {
     }
 
     public void delete(String relativePath) throws IOException {
-        Files.delete(getAssetsAbsolutePath(relativePath));
+        Path path = getAssetsAbsolutePath(relativePath);
```
wwelling, line 91: Doesn't IOException handle file not found when it tries to read it? Should the controller advice log this warning?

wwelling: When reading file from IO to response stream the exception should occur.

wwelling (commented):
Complexity increased with little gain to the knowledge status of the Vire application introducing much more potential misuse. A communication should be established before any file uploaded is deleted. If a submission is completed the communication should condition as the reason for the file deletion. That message should either go to the student or the dispaplinary board!

PR #1683, opened 2023-04-06
Issue 1599, 1601: Dates, such as graduation dates, are stored incorrectly.

### LINKED ISSUE #1599: Graduation dates sometimes stored improperly such that graduation date filters will not work properly
These happened for submissions created after the migration so it must be something that happened in Vireo4.  The dates were off only by a time zone.  The date is stored as 2022-12-01T05:0:0.000Z or similar.  The dates in error were off by one timezone.  Presumably the timezone is based on when the graduation date was selected.  If a December graduation date is selected sometime during Daylight Savings Time (spring to fall) then the time recorded is actually 1 hour closer to GMT than it will be in December, but Dec 1 at midnight a timezone offset from GMT by 5 hours is Nov 30 11pm for a timezone offset from GMT by 6 hours.
(more rarely some dates were actually off by a day, e.g. 2022-12-02T05:0:0:0.000Z but I don't have a theory for that yet)

### LINKED ISSUE #1601: Date storage for year/month dates
- [ ] Change date format to UTC
- [ ] Provide SQL to migrate existing dates
- [ ] Update application logic on front end and back end

kaladay:
Resolves #1599
Resolves #1601

There are several problems causing this situation.
1. AngularJS Datepicker Problems
2. FieldProfileDirective Datepicker Problems
3. Work-around Datepicker Problems
4. System Input Types have Invalid Regex Patterns
5. Off-By-1-Day Problems
6. Incorrect Validation on Java Side

There are also some notes on migrating data:
1. Migrating Validators
2. Migrating Bad Data

### AngularJS Datepicker Problems
The AngularJS is not processing the datepickerOptions at all. I've dived into the code and found it always uses full date and time with timezone and does conversion. Given the state of AngularJS I expect there to be no fixes for their design decisions here. The date formats are only being used for the label and not for the value.

### FieldProfileDirective Datepicker Problems
The Code setting the datepickerOptions is also flat out wrong, which further proves that these settings are being ignored. For a Year+Month format, such as "MMMM yyyy", the maxDate cannot be month because it has a year!

### Work-around Datepicker Problems
Given the situation with the problems surrounding the Datepicker not doing what we want I can see only two obviois solutions.
1. Fork and fix or improve the AngularJS Datepicker and related modules (and also maintain them).
2. Implement a work-around of some sort.

I believe that 2 is the better choice at this time.

My particular approach in two is to just correct the data after we got it. Because the value happens to be in a date time format, the value can be converted into what we want using the same exact filters via the Javascript date filter (via the AngularJS filter for this).

This introduces a problem in that Vireo uses a generalized design. There is no obvious way to distinguish one predicate from another when it comes to what a valid value structure is. However, the value of the predicate can be used to determine this. As these are all configurable I decided that the simplest solution is to offload the logic into an appConfig property. This reduces the need for code changes in the future should some new predicate need to be added.

I implemented to methods of matching the predicate:
1. start
2. exact

The first allows for matching a set of similar predicates, such as `dc.date.`. The second allows for exact matching.
There is a potential third option, called 'regex', but I saw no reason to implement that at this time and I consider such a feature to be out of scope.

### System Input Types have Invalid Regex Patterns
The system input types in the `SYSTEM_Input_Types.json` are flat out wrong.

For the two input types, `INPUT_DEGREEDATE` and `INPUT_DATETIME`, the validation regex is for a date time but the validation message is either "MMMM YYYY" or "MM/DD/YYYY".
There are two possible problems here, either the validation regex is correct or the validation message is correct.

I am assuming that the regex is incorrect.

I replaced the regex with much simpler checks rather than trying to handle all possible cases.
This is easier but increases the possibility of say a non-existent month name being used.

I would note that `INPUT_DATETIME` is refactored to `INPUT_DATE`.

### Off-By-1-Day Problems

I discovered that the off-by-1-day can be introduced due to a problem with the date selection and UTC.
This only happens when the date is empty (before a value is first assigned, such as on a new submission).

Steps to reproduce:
1. Create a new submission (via the student submission interface) and go do the page where the Defense Date can be selected (do not bother with changing any other fields).
2. Select April 30, 2023.
3. Notice that it now says April 29, 2023.
4. Select April 30, 2023, again.
5. Notice that is now say April 30, 2023.

I suspect this is happeningbecause the AngularJS Datepicker is using datetimes.
When no local time is explicitly set, it uses UTC.
This is then cast to the local time, resulting in the observed off by one.

The solution is to do the opposite of what we want.
Don't use UTC in the UI.
Instead, just leave the timezone to whatever the browser is set to so that there can be no conversion.
We are truncating the time anyway, and so having an incorrect timezone is irrelevant.

### Incorrect Validation on Java Side

The GraduationMonthValidator is using `ValidationUtility.MONTH_REGEX`, which is pattern matching to a numeric month.
The graduation month is a "MMMM YYYY" format to my understanding and so using that validator is incorrect.
I changed the code to use a custom validatorn pattern of "MMMM YYYY".

### Migrating Validators

```sql
update input_type set validation_pattern = '^\w+ \d{4}$' where name = 'INPUT_DEGREEDATE';
update input_type set validation_pattern = '^\d{4}-\d{1,2}-\d{1,2}$' where name = 'INPUT_DATETIME';
update input_type set name = 'INPUT_DATE' where name = 'INPUT_DATETIME';
```

### Migrating Bad Data

Migrate the Defense Date:
```sql
update field_value set value = regexp_replace(value, '(\d+)-(\d+)-(\d+)[T]{0,}.*', '\1-\2-\3')
  where id in (
    select distinct fv.id from field_value fv inner join field_predicate fp on fv.field_predicate_id = fp.id where fp.value = 'defense_date' and fv.value ~ '(\d+)-(\d+)-(\d+)T.*'
  );

update field_value set value = regexp_replace(value, '(\d{1,2})/(\d{1,2})/(\d{1,4})', '\3-\1-\2')
  where id in (
    select distinct fv.id from field_value fv inner join field_predicate fp on fv.field_predicate_id = fp.id where fp.value = 'defense_date' and fv.value ~ '(\d+)/(\d+)/(\d+)'
  );
```

Migrate the Graduation Semester regarding the date time data:
```sql
update field_value set value = regexp_replace(value, '(\d+)-01-(\d+)[T]{0,}.*', 'January \1')
  where id in (
    select distinct fv.id from field_value fv inner join field_predicate fp on fv.field_predicate_id = fp.id where fp.value = 'dc.date.issued' and fv.value ~ '(\d+)-(\d+)-(\d+)(T.*){0,1}'
  );

update field_value set value = regexp_replace(value, '(\d+)-02-(\d+)(T.*){0,1}', 'February \1')
  where id in (
    select distinct fv.id from field_value fv inner join field_predicate fp on fv.field_predicate_id = fp.id where fp.value = 'dc.date.issued' and fv.value ~ '(\d+)-(\d+)-(\d+)(T.*){0,1}'
  );

update field_value set value = regexp_replace(value, '(\d+)-03-(\d+)(T.*){0,1}', 'March \1')
  where id in (
    select distinct fv.id from field_value fv inner join field_predicate fp on fv.field_predicate_id = fp.id where fp.value = 'dc.date.issued' and fv.value ~ '(\d+)-(\d+)-(\d+)(T.*){0,1}'
  );

update field_value set value = regexp_replace(value, '(\d+)-04-(\d+)(T.*){0,1}', 'April \1')
  where id in (
    select distinct fv.id from field_value fv inner join field_predicate fp on fv.field_predicate_id = fp.id where fp.value = 'dc.date.issued' and fv.value ~ '(\d+)-(\d+)-(\d+)(T.*){0,1}'
  );

update field_value set value = regexp_replace(value, '(\d+)-05-(\d+)(T.*){0,1}', 'May \1')
  where id in (
    select distinct fv.id from field_value fv inner join field_predicate fp on fv.field_predicate_id = fp.id where fp.value = 'dc.date.issued' and fv.value ~ '(\d+)-(\d+)-(\d+)(T.*){0,1}'
  );

update field_value set value = regexp_replace(value, '(\d+)-06-(\d+)(T.*){0,1}', 'June \1')
  where id in (
    select distinct fv.id from field_value fv inner join field_predicate fp on fv.field_predicate_id = fp.id where fp.value = 'dc.date.issued' and fv.value ~ '(\d+)-(\d+)-(\d+)(T.*){0,1}'
  );

update field_value set value = regexp_replace(value, '(\d+)-07-(\d+)(T.*){0,1}', 'July \1')
  where id in (
    select distinct fv.id from field_value fv inner join field_predicate fp on fv.field_predicate_id = fp.id where fp.value = 'dc.date.issued' and fv.value ~ '(\d+)-(\d+)-(\d+)(T.*){0,1}'
  );

update field_value set value = regexp_replace(value, '(\d+)-08-(\d+)(T.*){0,1}', 'August \1')
  where id in (
    select distinct fv.id from field_value fv inner join field_predicate fp on fv.field_predicate_id = fp.id where fp.value = 'dc.date.issued' and fv.value ~ '(\d+)-(\d+)-(\d+)(T.*){0,1}'
  );

update field_value set value = regexp_replace(value, '(\d+)-09-(\d+)(T.*){0,1}', 'September \1')
  where id in (
    select distinct fv.id from field_value fv inner join field_predicate fp on fv.field_predicate_id = fp.id where fp.value = 'dc.date.issued' and fv.value ~ '(\d+)-(\d+)-(\d+)(T.*){0,1}'
  );

update field_value set value = regexp_replace(value, '(\d+)-10-(\d+)(T.*){0,1}', 'October \1')
  where id in (
    select distinct fv.id from field_value fv inner join field_predicate fp on fv.field_predicate_id = fp.id where fp.value = 'dc.date.issued' and fv.value ~ '(\d+)-(\d+)-(\d+)(T.*){0,1}'
  );

update field_value set value = regexp_replace(value, '(\d+)-11-(\d+)(T.*){0,1}', 'November \1')
  where id in (
    select distinct fv.id from field_value fv inner join field_predicate fp on fv.field_predicate_id = fp.id where fp.value = 'dc.date.issued' and fv.value ~ '(\d+)-(\d+)-(\d+)(T.*){0,1}'
  );

update field_value set value = regexp_replace(value, '(\d+)-12-(\d+)(T.*){0,1}', 'December \1')
  where id in (
    select distinct fv.id from field_value fv inner join field_predicate fp on fv.field_predicate_id = fp.id where fp.value = 'dc.date.issued' and fv.value ~ '(\d+)-(\d+)-(\d+)(T.*){0,1}'
  );
```

Migrate the Graduation Semester regarding the month/day/year data:
```sql
update field_value set value = regexp_replace(value, '01/(\d+)/(\d+)', 'January \2')
  where id in (
    select distinct fv.id from field_value fv inner join field_predicate fp on fv.field_predicate_id = fp.id where fp.value = 'dc.date.issued' and fv.value ~ '\d+/(\d+)/(\d+)'
  );

update field_value set value = regexp_replace(value, '02/(\d+)/(\d+)', 'February \2')
  where id in (
    select distinct fv.id from field_value fv inner join field_predicate fp on fv.field_predicate_id = fp.id where fp.value = 'dc.date.issued' and fv.value ~ '\d+/(\d+)/(\d+)'
  );

update field_value set value = regexp_replace(value, '03/(\d+)/(\d+)', 'March \2')
  where id in (
    select distinct fv.id from field_value fv inner join field_predicate fp on fv.field_predicate_id = fp.id where fp.value = 'dc.date.issued' and fv.value ~ '\d+/(\d+)/(\d+)'
  );

update field_value set value = regexp_replace(value, '04/(\d+)/(\d+)', 'April \2')
  where id in (
    select distinct fv.id from field_value fv inner join field_predicate fp on fv.field_predicate_id = fp.id where fp.value = 'dc.date.issued' and fv.value ~ '\d+/(\d+)/(\d+)'
  );

update field_value set value = regexp_replace(value, '05/(\d+)/(\d+)', 'May \2')
  where id in (
    select distinct fv.id from field_value fv inner join field_predicate fp on fv.field_predicate_id = fp.id where fp.value = 'dc.date.issued' and fv.value ~ '\d+/(\d+)/(\d+)'
  );

update field_value set value = regexp_replace(value, '06/(\d+)/(\d+)', 'June \2')
  where id in (
    select distinct fv.id from field_value fv inner join field_predicate fp on fv.field_predicate_id = fp.id where fp.value = 'dc.date.issued' and fv.value ~ '\d+/(\d+)/(\d+)'
  );

update field_value set value = regexp_replace(value, '07/(\d+)/(\d+)', 'July \2')
  where id in (
    select distinct fv.id from field_value fv inner join field_predicate fp on fv.field_predicate_id = fp.id where fp.value = 'dc.date.issued' and fv.value ~ '\d+/(\d+)/(\d+)'
  );

update field_value set value = regexp_replace(value, '08/(\d+)/(\d+)', 'August \2')
  where id in (
    select distinct fv.id from field_value fv inner join field_predicate fp on fv.field_predicate_id = fp.id where fp.value = 'dc.date.issued' and fv.value ~ '\d+/(\d+)/(\d+)'
  );

update field_value set value = regexp_replace(value, '09/(\d+)/(\d+)', 'September \2')
  where id in (
    select distinct fv.id from field_value fv inner join field_predicate fp on fv.field_predicate_id = fp.id where fp.value = 'dc.date.issued' and fv.value ~ '\d+/(\d+)/(\d+)'
  );

update field_value set value = regexp_replace(value, '10/(\d+)/(\d+)', 'October \2')
  where id in (
    select distinct fv.id from field_value fv inner join field_predicate fp on fv.field_predicate_id = fp.id where fp.value = 'dc.date.issued' and fv.value ~ '\d+/(\d+)/(\d+)'
  );

update field_value set value = regexp_replace(value, '11/(\d+)/(\d+)', 'November \2')
  where id in (
    select distinct fv.id from field_value fv inner join field_predicate fp on fv.field_predicate_id = fp.id where fp.value = 'dc.date.issued' and fv.value ~ '\d+/(\d+)/(\d+)'
  );

update field_value set value = regexp_replace(value, '12/(\d+)/(\d+)', 'December \2')
  where id in (
    select distinct fv.id from field_value fv inner join field_predicate fp on fv.field_predicate_id = fp.id where fp.value = 'dc.date.issued' and fv.value ~ '\d+/(\d+)/(\d+)'
  );
```


smutniak (approved):
Nicely done!  I set up TTU's data locally and ran the scripts and tested.  I then ran the scripts a second time to make sure there were no further changes.

cstarcher (changes requested):
I received the following error when filtering by Approval Date, Committee Approval Date, Committee Embargo Approval Date, License Agreement Date, and Submission Date. The date shows under Now Filtering By, but the list is not filtered.

Screenshot 2023-04-12 at 8 21 02 AM

I just realized that the above error did not include the the actual exception.

Caused by: org.h2.jdbc.JdbcSQLSyntaxErrorException: Column "PFV8.VALUE" not found; SQL statement:

cstarcher (approved):
#1696 resolves errors. Everything else looks good.

kaladay:
Belated follow up.
Some dates are in this form: `2015-10-16 00:00:00`.

A script above has an incorrect `fv.value ~ ...` condition that results in these being missed.

This query should work:
```sql
update field_value set value = regexp_replace(value, '(\d+)-(\d+)-(\d+)[T]{0,}.*', '\1-\2-\3')
  where id in (
    select distinct fv.id from field_value fv inner join field_predicate fp on fv.field_predicate_id = fp.id where fp.value = 'defense_date' and fv.value ~ '(\d+)-(\d+)-(\d+)[T]{0,}.*'
  );
```

PR #1685, opened 2023-04-11
Fix reordering response handling overlapping with broadcast listen

wwelling:
May need to disable drag and drop until ready again.


src/main/webapp/app/controllers/settings/emailTemplateRepoController.js
```diff
@@ -120,14 +120,13 @@ vireo.controller("EmailTemplateRepoController", function ($controller, $scope, $
             container: '#email-templates'
         });
 
-        EmailTemplateRepo.listen(function (data) {
-            $scope.resetEmailTemplates();
+        EmailTemplateRepo.listen(function (/* data */) {
```
kaladay, line 123: I think that it would be ideal to remove the `/* data */` rather than leave unused commented out code.

src/main/webapp/app/controllers/settings/emailTemplateRepoController.js
```diff
@@ -120,14 +120,13 @@ vireo.controller("EmailTemplateRepoController", function ($controller, $scope, $
...
+            $timeout(function () {
+                EmailTemplateRepo.reset().then(function () {
+                    $scope.resetEmailTemplates();;
```
kaladay, line 126: One too many semi-colons.
Can you remove the extra one?

kaladay (changes requested):

kaladay (approved):

PR #1686, opened 2023-04-11
Style submission list column sort icons

wwelling:
Opted to display muted icons until column sorted.

image


kaladay (approved):
We really do need something, like a "loading", a "busy", or locking the button.
That need is out of scope to this issue.
This PR looks good and it runs as described.

PR #1687, opened 2023-04-11
1672 input placeholder

### LINKED ISSUE #1672: Pre-populated inputs
Relevant Components:

- All Inputs

Description
Best practices suggest that inputs not be populated with filler text on page load. The labels will provide the information necessary, and the reminder text will provide it again. There is no need for the filler text, and it causes two particular challenges.

First, it is redundant, which causes an increase in cognitive load, especially for users who cannot see the element. Currently, the way the element is coded, assistive technology will announce the purpose of the input, then announce that there is an input, then announce the filler text (the purpose of the input again).

Second, it causes more work for every user. Generally, it is expected that a user should navigate to the input and start typing. In this case, something must be done about the text that already exists inside the input.

Solution
Best practice: Take out the filler text.

cstarcher:
Resolves #1672


kaladay:
There is an input field that is using placeholder text as a value rather than as a placeholder.

Screenshot of "Choose Your Organization" placeholder

This appears to be populated using `new_submission_header_text` here:
- https://github.com/TexasDigitalLibrary/Vireo/blob/dff6a55c064205c64574431e70169dc10f0fea01/src/main/resources/settings/SYSTEM_Defaults.json#L25

I would note that I do not know if this is proper or not off the top of my mind but I thought it to be worth mentioning.

kaladay (approved):

cstarcher:
@kaladay That is intended to be a default value for the field. The value is not on a placeholder attribute. It is actually on the model attribute for validatedinput.

https://github.com/TexasDigitalLibrary/Vireo/blob/dff6a55c064205c64574431e70169dc10f0fea01/src/main/webapp/app/views/admin/settings/application/organizationSettings.html#L23

PR #1698, opened 2023-04-14
set proper contrast throughout application

cstarcher:
Set proper contrast throughout application for all background and foreground elements.

Added configurations for color selections.

### DIFF
--- a/src/main/webapp/app/views/admin/settings/application/lookAndFeel.html
+++ b/src/main/webapp/app/views/admin/settings/application/lookAndFeel.html
@@ -3,15 +3,51 @@
 		<div class="row">
 			<div class="col-sm-6 col-lg-9">
 
-				<div class="row">
-					<h4>Background
-					   <div tooltip="Use hex color values to change the background on the student side. The main color (default #1B333F) fades into the highlight color (default #43606E).">
-					   	<span class="glyphicon glyphicon-info-sign opaque color-reset"></span>
-					   </div>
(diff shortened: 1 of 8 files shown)


src/main/webapp/app/views/admin/settings/application/lookAndFeel.html
```diff
@@ -3,15 +3,51 @@
...
+						<div class="input-group color-selector">
+							<span class="input-group-addon">#</span>
+						  	<input
+						  		class="form-control color-input"
+						  		ng-change="updateHexValue('lookAndFeel','text_main_color')"
+						  		ng-model="settings.configurable.lookAndFeel.text_main_color.value"
+						  		dehashcolor>
+
+							<span class="input-group-addon">
+								<label class="color-selector-swatch" ng-style="{background: settings.configurable.lookAndFeel.text_main_color.value}">
```
kaladay, line 32: Would it be better to use a custom class and have that class setting the background added here rather than using `ng-style`?

kaladay (approved):
Approving with the hopes of another issue being created for supporting indepedent text color controls for both main and highlight background colors.

src/main/webapp/app/views/admin/settings/application/lookAndFeel.html
kaladay: Okay, so this is existing design choice behavior.

PR #1703, opened 2023-04-17
Check if refresh on model is a function

wwelling:
Resolves a regression introduced by https://github.com/TAMULib/Weaver-UI-Core/pull/232 where keydown escape event handler is invoked when a $scope.model.refresh method is not a function.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "build": "wvr build --clean"
   },
   "dependencies": {
-    "@wvr/core": "2.2.4",
+    "@wvr/core": "https://github.com/TAMULib/Weaver-UI-Core.git#modal-keyboard-escape-fix",
     "angular-ui-tinymce": "0.0.19",
     "file-saver": "2.0.5",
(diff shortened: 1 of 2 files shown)


cstarcher (approved):
This fixes the three related issues. One issue I found while testing was some strange behavior  in modals with hidden elements (e.g. Add File). However, I think that is out of the scope of these issues and can be addressed separately.

package.json
```diff
@@ -22,7 +22,7 @@
     "build": "wvr build --clean"
   },
   "dependencies": {
-    "@wvr/core": "2.2.4",
+    "@wvr/core": "https://github.com/TAMULib/Weaver-UI-Core.git#modal-keyboard-escape-fix",
```
kaladay, line 25: Be sure to revert/change this once weaver-ui-core is updated to the supporting weaver-ui-core release.

wwelling:
Closing to recreate PR.

PR #1718, opened 2023-04-24
Issue 1669: Make field profile input info addon tab accessible

wwelling:


kaladay (approved):
This adds tabIndex as requested by the issue.

PR #1722, opened 2023-04-24
Issue 1714: Fix customize filters modal drag and drop behavior

wwelling:
The angular filter in the template was causing the index being offset. Moving the filter into the controller before the template render resolved the issue. Additional refactoring and formatting.


cstarcher (commented):
Filter are arranging correctly, but they do not save when clicking the Save button.

wwelling:
@cstarcher thanks for the review. Will resolve ASAP.

wwelling:
@cstarcher I see. The save actually does persist but reverts changes on response to request. You should see the intended custom filters and order on page refresh. Fixing the response handling now.

wwelling:
@cstarcher, I believe I have the issue resolved now. Please also test customize view along with customize filters. Both should work with and without page refresh.

PR #1725, opened 2023-04-24
Issue 1671: Open date picker on enter

wwelling:
Additional minor formatting cleanup.


cstarcher (approved):
Calendar (datepicker) elements are now accessible by using the arrow keys to navigate, the enter key to select, and the esc key to exit.

PR #1726, opened 2023-04-25
Issue 1582: Reset list view pagination and fix exposed pagination related problems.

### LINKED ISSUE #1582: view must be reset before filter applied lest no items show in list
The default sort order for the list view is “ID ascending”, and I have my default view set up to show 100 rows per page. I will periodically move to the end of the list (i.e., last page) to see the latest submissions. (See 1st screenshot for an example.)

If I then apply a filter, the display tells me how many records are in the filter (e.g., 22), but I don’t see the items in the list view, nor do I see any page numbers to pick from. (See 2nd screenshot for an example.)

To resolve the problem, I have to clear the filter, reset my view to be on the 1st page, and then reapply the filter to see the filtered items. And this would appear to be extra, unnecessary steps.
- Jon C.

kaladay:
resolves #1582 

Change the behavior to reset the pagination when a filter is applied.
This prevents the bug where the list is empty and the table is unusable when a filter is applied.

Making this change exposed several related problems and bugs that have been deemed necessary to be fixed as part of this issue.

A new `resetPagination()` function is created that performs the reset task.
This required replacing duplicated code that also performs the pagination.
This change exposed a bug where the `update()` function is always calling the database list view load even when unnecessary.

This additional list load request in `update()` ends up altering the pagination and must be fixed to get this new behavior to work consistently.
Preventing it from reloading the list when not needed also improves performance by removing an extra unnecessary request in certain places.
Break apart the `update()` function, adding an additional function `processUpdate()`.
This function performs the tasks needed.
Breaking this up is required because there is an additional request to `ManagerSubmissionListColumnRepo.submissionListPageSize()` that alters the page size.
The `ManagerSubmissionListColumnRepo.submissionListPageSize()` effectively reset the selected page count, causing a regression.
The new behavior is to only call `ManagerSubmissionListColumnRepo.submissionListPageSize()` when there is no local session data.
This design change maintains pre-existing behavior with the page count.
This is another likely performance improvement where the page count is no longer fetched when not necessary.

Despite the improvements, I still found a case where the page list is empty and the table is broken.
An example would be data changes since pagination is set.
It turns out that all of the data needed to determine if the selected pagination is correct is returned on the list response.
The problem is that the list response is already loaded so even if the pagination data were to be corrected there is no row data to display.
This is solved by making an extra request with the pagination data corrected using data from the first request.
Recursion is completely undesirable and so this extra request is performed exactly once and then only in the specific situation where the first request is incorrect.
Doing this required breaking apart the `queryGetData()` and adding a new parameter for forcing the page number.

Removing a single row from the list view should not reset the pagination, unlike the other filters.
When a filter that removes a single row is restored then the pagination must also not be reset.
This preserves existing expected behavior.
This exposed another bug is observed where when the pagination is restored page index may be incorrect.
This is because the database stores the page numbers starting at 0 while the UI table is storing page numbers starting at 1.
There are some cases where 1 is incorrectly added or not added.
Fixing this ended up being necessary as part of solving the original issue requested to be fixed.
The solution here is that when the value comes from the database, increment it by one.
The value is preserved, unchanged, in all other cases.

When removing a single row and that row is the last row for the page then the pagination must be updated.
Change the current page number to the previous page in the case where the current page is not the first page (page index 0 as per DB or page index 1 as per UI).


wwelling:
https://github.com/TexasDigitalLibrary/Vireo/issues/1582#issuecomment-1515052915

kaladay:
This change doesn't work in isolation.

Just as I described in the description.

wwelling:
Is there a solution in which does work in isolation?

wwelling:
> Change the behavior to reset the pagination when a filter is applied. This prevents the bug where the list is empty and the table is unusable when a filter is applied.
> 
> Making this change exposed several related problems and bugs that have been deemed necessary to be fixed as part of this issue.
> 
> A new `resetPagination()` function is created that performs the reset task. This required replacing duplicated code that also performs the pagination. This change exposed a bug where the `update()` function is always calling the database list view load even when unnecessary.
> 
> This additional list load request in `update()` ends up altering the pagination and must be fixed to get this new behavior to work consistently. Preventing it from reloading the list when not needed also improves performance by removing an extra unnecessary request in certain places. Break apart the `update()` function, adding an additional function `processUpdate()`. This function performs the tasks needed. Breaking this up is required because there is an additional request to `ManagerSubmissionListColumnRepo.submissionListPageSize()` that alters the page size. The `ManagerSubmissionListColumnRepo.submissionListPageSize()` effectively reset the selected page count, causing a regression. The new behavior is to only call `ManagerSubmissionListColumnRepo.submissionListPageSize()` when there is no local session data. This design change maintains pre-existing behavior with the page count. This is another likely performance improvement where the page count is no longer fetched when not necessary.
> 
> Despite the improvements, I still found a case where the page list is empty and the table is broken. An example would be data changes since pagination is set. It turns out that all of the data needed to determine if the selected pagination is correct is returned on the list response. The problem is that the list response is already loaded so even if the pagination data were to be corrected there is no row data to display. This is solved by making an extra request with the pagination data corrected using data from the first request. Recursion is completely undesirable and so this extra request is performed exactly once and then only in the specific situation where the first request is incorrect. Doing this required breaking apart the `queryGetData()` and adding a new parameter for forcing the page number.
> 
> Removing a single row from the list view should not reset the pagination, unlike the other filters. When a filter that removes a single row is restored then the pagination must also not be reset. This preserves existing expected behavior. This exposed another bug is observed where when the pagination is restored page index may be incorrect. This is because the database stores the page numbers starting at 0 while the UI table is storing page numbers starting at 1. There are some cases where 1 is incorrectly added or not added. Fixing this ended up being necessary as part of solving the original issue requested to be fixed. The solution here is that when the value comes from the database, increment it by one. The value is preserved, unchanged, in all other cases.
> 
> When removing a single row and that row is the last row for the page then the pagination must be updated. Change the current page number to the previous page in the case where the current page is not the first page (page index 0 as per DB or page index 1 as per UI).

Could you help interpret this text with comments inline to the diff of the PR?

wwelling:
Should we consider not prematurely optimizing?

Whenever a filter is applied (added/removed/modified), the page number set to 0 (or 1 whichever is the initial index) and issue query. Cleaning the code is one thing but changing the logic is another. And yes, the code certainly could use some cleanup. However, additional logic tends to make it messier if not cleaned up adequately.

src/main/webapp/app/controllers/submission/submissionListController.js
```diff
@@ -29,6 +29,8 @@ vireo.controller("SubmissionListController", function (NgTableParams, $controlle
 
     $scope.fieldPredicates = FieldPredicateRepo.getAll();
 
+    var rowFilterTitle = "Exclude";
```
wwelling, line 32: Do we need this?

src/main/webapp/app/controllers/submission/submissionListController.js
```diff
@@ -51,85 +53,142 @@ vireo.controller("SubmissionListController", function (NgTableParams, $controlle
...
-                count: $scope.page.count
+                page: pageNumber,
+                count: pageCount
             }, {
                 counts: $scope.page.options,
                 total: $scope.page.totalElements,
                 filterDelay: 0,
                 getData: function (params) {
                     start = window.performance.now();
-                    return SubmissionRepo.query($scope.userColumns, params.page() > 0 ? params.page() - 1 : params.page(), params.count()).then(function (response) {
```
wwelling, line 63: What is the purpose of moving this?

src/main/webapp/app/controllers/submission/submissionListController.js
```diff
@@ -51,85 +53,142 @@ vireo.controller("SubmissionListController", function (NgTableParams, $controlle
         var start;
 
         var query = function () {
+            var pageNumber = $scope.page.number;
+            var pageCount = $scope.page.count;
+
+            if (angular.isDefined(sessionStorage.getItem("list-page-number")) && sessionStorage.getItem("list-page-number") !== null) {
```
wwelling, line 59: Why not check session storage once for each?

Better yet, why check session storage each time the query is issued?

Shouldn't the scope already be updated?

wwelling:
We should see if this PR works correctly with https://github.com/TexasDigitalLibrary/Vireo/pull/1728

src/main/webapp/app/controllers/submission/submissionListController.js
kaladay: We already do this in the pre-existing code.
We use the same string in multiple places now so I changed the scope appropriately.

src/main/webapp/app/controllers/submission/submissionListController.js
kaladay: As described in the PR description:

> Despite the improvements, I still found a case where the page list is empty and the table is broken.
An example would be data changes since pagination is set.
It turns out that all of the data needed to determine if the selected pagination is correct is returned on the list response.
The problem is that the list response is already loaded so even if the pagination data were to be corrected there is no row data to display.
This is solved by making an extra request with the pagination data corrected using data from the first request.
Recursion is completely undesirable and so this extra request is performed exactly once and then only in the specific situation where the first request is incorrect.
Doing this required breaking apart the queryGetData() and adding a new parameter for forcing the page number.

src/main/webapp/app/controllers/submission/submissionListController.js
kaladay: A simple oversight.
I can switch to a variable:
```js
   var pageNumber = sessionStorage.getItem("list-page-number");
```

> Shouldn't the scope already be updated?

That is session storage data and may be updated outside of the view.

wwelling (approved):

cstarcher (approved):

wwelling (changes requested):
I prefer https://github.com/TexasDigitalLibrary/Vireo/pull/1758

wwelling:
> Making this change exposed several related problems and bugs that have been deemed necessary to be fixed as part of this issue.

Are we certain beyond any reasonable doubt that these changes didn't cause problems and bugs? This is irrelevant now that there is a solution provided that does not change other behavior besides the requested change of reset page number when filtering is modified.

Could we please test the alternative solution that is resolved the issue and does not have any additional regressions?

Issue #1667, opened 2023-03-30
Using a Saved Filter can cause system to be unresponsive for long periods of time.


kaladay:
### The Problem
After having customizing a view and thensaving the filters, I selected the saved filter.
After loading I may have refreshed the page.
The page just sat their looking broken.
Several long seconds later, the page suddenly worked.

### To Reproduce
1. Add some columns to the Customize View, such as Custom Actions and Notes.
2. Select Title, Customize Filter, and add Custom Actions and Notes as well.
3. Filter by a Title, such as "test" (Make sure "test" is set for Title on at least one Submission).
4. Select "Save current filter".
  1. Fill in the name, such as "Filter by Test Title".
  2. Select the Public checkbox of "I would like this filter to be available to all users.".
  3. Select the Columns checkbox of "I would like this filter to include visible columns.".
  4. Click Save.
5. Under the "Saved Filters", under "With Columns", Select the saved filter by the name you just created, such as "Filter by Test Title".
6. Seems to work, now refresh the page.
7. The problem is observed here to have a very slow load that does not seem to work. Observe the network log.

cstarcher:
Decomposition and actionable cards for resolution.

wwelling:
*then saving

Unable to reproduce.

@kaladay How many submissions did you have generated? Was expansive enabled for all of them?

kaladay:
I had 111000 Submissions at the demo where this problem was clearly observed.
Expansive was enabled.

I filtered by "title" with a value of "test" and saved that filter.
Compare filtering by "title" without using the saved filter vs filtering by "title" via the saved filter. A huge difference.

It can also help to have multuple browsers/tabs open to refresh and see the problem get worse fast.

wwelling:
What do you mean? `Select Title, Customize Filter, and add Custom Actions and Notes as well.`

wwelling:
Do you mean add a title filter?

wwelling:
>Filter by a Title, such as "test" (Make sure "test" is set for Title on at least one Submission).
generated data test is in all title. not relevant experiment to reproduce ... induced so far.

wwelling:
>Select the Public checkbox of "I would like this filter to be available to all users.".
Select the Columns checkbox of "I would like this filter to include visible columns.".

select is the incorrect verb. it should be check the public and column checkbox

wwelling:
[...]

PR #1728, opened 2023-04-25
Issue 1667: Improved submission list view performance

wwelling:
Switching the `ManyToMany` join between `NamedSearchFilterGroup` and `SubmissionListColumn` to `LAZY` fetch mode improved submission list view page load time by a factor of 2 when applying filter described in [Using a Saved Filter can cause system to be unresponsive for long periods of time.#1667](https://github.com/TexasDigitalLibrary/Vireo/issues/1667) with small data.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -115,7 +115,7 @@
     <dependency>
       <groupId>com.googlecode.libphonenumber</groupId>
       <artifactId>libphonenumber</artifactId>
-      <version>8.13.7</version>
+      <version>8.13.8</version>
     </dependency>
 
     <dependency>
@@ -641,7 +641,7 @@
                     </goals>
                   </pluginExecutionFilter>
                   <action>
-                    <execute />
+                    <execute/>
                   </action>
                 </pluginExecution>
               </pluginExecutions>
@@ -675,7 +675,7 @@
         </os>
       </activation>
       <properties>
-        <npm-extension></npm-extension>
+        <npm-extension/>
       </properties>
     </profile>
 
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -22,7 +22,7 @@
     "build": "wvr build --clean"
   },
   "dependencies": {
-    "@wvr/core": "2.2.5-rc.2",
+    "@wvr/core": "https://github.com/TAMULib/Weaver-UI-Core.git#piggy-back-poorly-scoped-requests",
```
kaladay, line 25: Could you revert this change?
I don't think we should commit github URLs to staging, etc...

kaladay (changes requested):
I've investigated the problem and compared the before and after with this PR.

I believe this does not address the problem.

I can observe the following:

Searching for title, such as "test", results in say 800ms to 1000ms for my particular result set.

Using the saved filter that is the same search as "test" above results in a 2000ms to 3000ms delay.

Using the unsaved filter versus using the saved filter, both with the same search results, should have roughly same performance rather than one being more than twice as slow as the other.

I cannot say that this PR solves the problem described in the referenced issue, however, this PR might still be good as a general performance optimization.

package.json
wwelling: That version of `@wvr/core` is unrelated to the optimization for the steps your provided. I will gladly remove and wait further review there.

kaladay (approved):

kaladay:
This message is for reference purposes.

The changes in this PR were re-introduced here: https://github.com/TexasDigitalLibrary/Vireo/pull/1749
(which is now merged.)

PR #1732, opened 2023-04-26
added debounce for colorpicker

cstarcher:


wwelling (approved):
https://docs.angularjs.org/api/ng/directive/ngModelOptions

Syntax looks correct.

PR #1753, opened 2023-05-01
Merger for PR 1751 for helping resolve 1667.

kaladay:
merger for #1751

I need to do some testing to make sure the merge didn't introduce any problems.
I will add a comment once I confirm/deny this on my end of things.


wwelling:
@kaladay could you please provide a few before and after SQL examples? Please try to include a good variation of filters and columns. At least a base case of default filters and columns and maximum edge case of all filters and columns.

PR #1755, opened 2023-05-01
4.2.0 Docker update replacing deprecated base images

wwelling:
- Replace base image `maven:3-openjdk-11-slim` with `maven:3-eclipse-temurin-11-alpine`.
- Replace base image `openjdk:11-jre-slim` with `eclipse-temurin:11-alpine`.
- Add [Adminer ](https://www.adminer.org/) to docker compose.
- Use named volumes with docker compose.
- Minor cleanup of example env file.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -80,6 +80,7 @@ RUN apk -U upgrade && \
 # Copy files from outside docker to inside.
 COPY build/appConfig.js.template /usr/local/app/templates/appConfig.js.template
 COPY build/docker-entrypoint.sh /usr/local/bin/docker-entrypoint.sh
+RUN chmod ugo+x /usr/local/bin/docker-entrypoint.sh
 
 # Login as user.
 USER $USER_NAME
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,36 +1,36 @@
-# Settings.
+# Build arguments.
 ARG USER_ID=3001
 ARG USER_NAME=vireo
-ARG SOURCE_DIR=/$USER_NAME/source
+ARG HOME_DIR=/$USER_NAME
+ARG SOURCE_DIR=$HOME_DIR/source
+ARG APP_PATH=/var/vireo
 ARG NODE_ENV=production
 
 # Maven stage.
-FROM maven:3-openjdk-11-slim as maven
+FROM maven:3-eclipse-temurin-11-alpine as maven
 ARG USER_ID
 ARG USER_NAME
+ARG HOME_DIR
 ARG SOURCE_DIR
+ARG APP_PATH
 ARG NODE_ENV
 
 ENV NODE_ENV=$NODE_ENV
 
-# Create the user and group (use a high ID to attempt to avoid conflicts).
-RUN groupadd --non-unique -g $USER_ID $USER_NAME && \
-    useradd --non-unique -d /$USER_NAME -m -u $USER_ID -g $USER_ID $USER_NAME
+# Create the group (use a high ID to attempt to avoid conflits).
+RUN addgroup -g $USER_ID $USER_NAME
 
-# Install stable Nodejs and npm.
-RUN apt-get update && \
-    apt-get upgrade -y && \
-    apt-get install -y nodejs npm iproute2 && \
-    apt-get clean && \
-    rm -rf /var/lib/apt/lists/* && \
-    npm cache clean -f && \
-    npm install -g n && \
-    n stable
+# Create the user (use a high ID to attempt to avoid conflits).
+RUN adduser -h $HOME_DIR -u $USER_ID -G $USER_NAME -D $USER_NAME
 
 # Ensure source directory exists and has appropriate file permissions.
 RUN mkdir -p $SOURCE_DIR && \
     chown $USER_ID:$USER_ID $SOURCE_DIR && \
     chmod g+s $SOURCE_DIR
+
+# Upgrade the system and install dependencies.
+RUN apk -U upgrade && \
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -1,37 +1,37 @@
...
+ARG APP_PATH
 ARG NODE_ENV
 
 ENV NODE_ENV=$NODE_ENV
 
-# Create the user and group (use a high ID to attempt to avoid conflicts).
-RUN groupadd --non-unique -g $USER_ID $USER_NAME && \
-    useradd --non-unique -d /$USER_NAME -m -u $USER_ID -g $USER_ID $USER_NAME
+# Create the group (use a high ID to attempt to avoid conflits).
+RUN addgroup -g $USER_ID $USER_NAME
```
kaladay, line 21: What is the reason for removing `--non-unique`?
This was done in other repositories to ensure this does not fail.

Also, this `addgroup` and that `adduser` line below were combined for performance reasons. Are we throwing this performance reason away now? Why? Is the a bug in regards to this?

Dockerfile
```diff
@@ -1,37 +1,37 @@
...
+RUN adduser -h $HOME_DIR -u $USER_ID -G $USER_NAME -D $USER_NAME
 
 # Ensure source directory exists and has appropriate file permissions.
 RUN mkdir -p $SOURCE_DIR && \
     chown $USER_ID:$USER_ID $SOURCE_DIR && \
     chmod g+s $SOURCE_DIR
 
+# Upgrade the system and install dependencies.
+RUN apk -U upgrade && \
+    apk add --update --no-cache nodejs npm make g++ py3-pip
```
kaladay, line 33: I thought all of that cache clean code in the old apt-get update was necessary.
Is `apk` different enough that this is not a problem?
Or perhaps, it just doesn't support that functionality?

Dockerfile
```diff
@@ -52,48 +52,46 @@ USER $USER_NAME
...
-# Create the user and group (use a high ID to attempt to avoid conflicts).
-RUN groupadd --non-unique -g $USER_ID $USER_NAME && \
-    useradd --non-unique -d /$USER_NAME -m -u $USER_ID -g $USER_ID $USER_NAME
-
 # Login as user.
 USER $USER_NAME
 
 # Set deployment directory.
-WORKDIR /$USER_NAME
+WORKDIR $HOME_DIR
```
kaladay, line 88: It might be worth considering using `$HOME_DIR` rather than `/$USER_NAME` for othe docker files as well.

Dockerfile
wwelling: Pending the base image. The updates over time become a vulnerability.

Dockerfile
wwelling: Possibly. Did it startup on your development environment or are you still taking a piss?

Dockerfile
kaladay: I believe `--non-unique` is added here:
- https://github.com/TAMULib/CAP/pull/186

I believe the reason was because of possible existing conflicts in the docker system.

For that `groupadd` and `useradd` grouping them together is what you had advocated for, for performance reasons. I simple have not objected.

Dockerfile
kaladay: I am asking about `apk` usage here.
Do you know that `apk` does not need to cache changes that `apt-get` does?

Dockerfile
kaladay: Haven't tested this yet.
I am merely remarking that we might want to apply this same change to Dockerfiles in our other repositories.

kaladay (changes requested):
I get
```
ERROR: Version in "./docker-compose.yml" is unsupported. 
```
So I changed the version back to `3.7` and docker compose is running.

Once it started running, I ended up getting this error:
```
successfully tagged 127.0.0.1/tdl/vireo:4.2.0
WARNING: Image for service vireo was built because it did not already exist. To rebuild this image you must use `docker-compose build` or `docker-compose up --build`.
Creating vireo_db_1 ... done
Creating vireo_adminer_1 ... done
Creating vireo           ... error

ERROR: for vireo  Cannot start service vireo: OCI runtime create failed: container_linux.go:367: starting container process caused: exec: "docker-entrypoint.sh": executable file not found in $PATH: unknown

ERROR: for vireo  Cannot start service vireo: OCI runtime create failed: container_linux.go:367: starting container process caused: exec: "docker-entrypoint.sh": executable file not found in $PATH: unknown
ERROR: Encountered errors while bringing up the project.
```

~The `docker-entrypoint.sh` needs to be directly called or added to one of the paths, such as `/usr/local/bin`.~

Correction, it looks like it is copied there.
It probably needs to have the executable bit set.
This should be done in the Dockerfile after copying the file over.
Something like:
`chmod ugo+x /usr/local/bin/docker-entrypoint.sh`

Dockerfile
```diff
@@ -52,48 +52,46 @@ USER $USER_NAME
...
+    chown $USER_ID:$USER_ID $APP_PATH && \
+    chmod g+s $APP_PATH
+
+# Update the system and install gettext for envsubst.
+RUN apk -U upgrade && \
+    apk add --update --no-cache gettext
 
 # Copy files from outside docker to inside.
 COPY build/appConfig.js.template /usr/local/app/templates/appConfig.js.template
 COPY build/docker-entrypoint.sh /usr/local/bin/docker-entrypoint.sh
```
kaladay, line 83:
```suggestion
RUN chmod ugo+x /usr/local/bin/docker-entrypoint.sh

```

kaladay (approved):

PR #1760, opened 2023-05-03
Issue 1743: Add submission list view customize column and default column buttons

wwelling:
Ensure successful response is handled in https://github.com/TexasDigitalLibrary/Vireo/pull/1758


cstarcher (changes requested):
resetColumnsToDefault should be called in the removeFilterValue function (line 637) as well. This function is called when removing the filter by clicking the glyphicon-remove-circle next to the filter name under Now Filtering By.

wwelling:
@cstarcher are we sure we want to do that by removing a single filter from the named search filter group?


When `Clear current filter` is clicked the users active filter (named search filter group) is cleared/reset. See https://github.com/TexasDigitalLibrary/Vireo/blob/main/src/main/java/org/tdl/vireo/controller/SubmissionListController.java#L266.

When removing an individual applied filter from `Now Filtering By:` it is surgically removed from the active filter (named search filter group). See https://github.com/TexasDigitalLibrary/Vireo/blob/main/src/main/java/org/tdl/vireo/controller/SubmissionListController.java#L231.

Isn't intended functionality for ability to add/remove/reorder "filter forms", applied filters, and columns for the current users active filter? If so, I do not think the columns should be reset to default when removing an individual applied filter.

cstarcher:
@wwelling You're correct. I was thinking about this in relation to saved filters. 

I still think this is not ideal. What if there were a check for savedColumns on the namedSearchFilterGroup? If they exist, call resetColumnsToDefault. Otherwise, just removeFilter.

wwelling:
Perhaps we should move reset columns above the table next to customize view?

I thought a users active filter is meant to be mutable without saving it and selecting public and include columns.

What use case is there for clearing all columns when an individual filter is removed? I feel coupling remove individual filter with resetting columns with make for a restrictive UI/UX when creating an active filter to save and share.

Imagine spending 30 minutes arranging the columns and then removing a single applied filter and all the work is lost without saving current filter?

wwelling:
I think we may be missing update names search filter group functionality in the UI. There is a disconnect between a users active filter and saved filter. Also, if we are to adopt the concurrent modification paradigm we will need some adjustments to the UI and back end.

wwelling:
> I still think this is not ideal. What if there were a check for savedColumns on the namedSearchFilterGroup? If they exist, call resetColumnsToDefault. Otherwise, just removeFilter.

Something like this could be sufficient to determine whether the columns should be reset when an applied filter is removed.

However, when an user modifies their active filter it is no longer a "saved" filter.

wwelling:
Merge branch not worth the effort here.

cstarcher (approved):
This looks really good! I think this was a much better solution than using the clear current filter.

kaladay (approved):

PR #1762, opened 2023-05-03
Issue 1712: Use anonymous download from weaver file service

wwelling:
Additional use link styled span to avoid multiple event handlers.

Requires https://github.com/TAMULib/Weaver-UI-Core/pull/234

`@wvr/core` dependency will be updated upon next patch release.


package.json
```diff
@@ -22,7 +22,7 @@
     "build": "wvr build --clean"
   },
   "dependencies": {
-    "@wvr/core": "2.2.4",
+    "@wvr/core": "git@github.com:TAMULib/Weaver-UI-Core.git#wwelling-patch-1",
```
kaladay, line 25: Please do not add the github URL dependency in the PR going into staging.
Noting that this PR requires a weaver update should be good enough.

kaladay (changes requested):

package.json
wwelling: This is to afford testing of both Vireo and Weaver without additional effort from the reviewer.

wwelling: Please do not make this more difficult than it needs to be.

kaladay (approved):

cstarcher (approved):

jcreel (approved):

PR #1765, opened 2023-05-04
Issue 1706: Remove the auto-focus from the submission.

### LINKED ISSUE #1706: First form field is auto-focused, which makes it unlikely a screen reader user will discover the form instructions above

kaladay:
resolves #1706

There are two focus directives that each play a part in the auto-focus.
Removing them both from the submission form solves the problem.
Completely remove the directives and tests because they are not used anywhere else.

The auto-focus directive is only used on the submission view.
The field profile directive may be used in more places.
The field profile focus directive very clearly is causing the focus problem.


wwelling (approved):
Does what the issue asks for. Thanks for the cleanup.

cstarcher (changes requested):
I should have been more explicit when reporting 1706. Even so, I'm not sure this resolves the issue. As stated in the title, the concern is that the "Instructions" text above the input fields is not accessible to a screenreader. This does move the focus to the submission step buttons which is preferable to the first input field. However, the "Instructions" text is still not accessible using VoiceOver on MacOS.

kaladay:
> Can someone confirm that this does work for any other screen reader?
The https://getpericles.com screenreader does read the instructions in my Firefox browser.

cstarcher (approved):

PR #1793, opened 2023-05-26
4.2.1 staging

kaladay:


src/test/java/org/tdl/vireo/model/DocumentTypeTest.java
```diff
@@ -56,6 +62,8 @@ public void testCascade() {
     }
 
     @Test
+    @Disabled // FIXME: This function's name suggested testing the delete, but all this does is call create.
```
kaladay, line 65: I didn't comment this one correctly.
Inside of the assertThrows(), the delete is being performed.

No exception is thrown anymore.
The delete is called but the delete doesn't actually happen as if only the create is called.

```
DocumentTypeTest.testFieldPredicateDeleteFailure:68 Expected org.springframework.dao.DataIntegrityViolationException to be thrown, but nothing was thrown
```

cstarcher (approved):

Issue #1756, opened 2023-05-02
Submission choose your organization loading slowly


cstarcher:
When selecting “Start your submission” after logging in, the lag for the organization list to appear is considerable. It only happens on the initial load. Once it is loaded, you can go away and come back to it quickly within the same session.

This also shows up when a user has pending submissions. There is a significant lag for the Start your submission button to change to Manage/View your submission(s).

Examining network traffic indicates that the slowdown is occurring with these requests. The latter being the worst culprit.

/organization/all
/submission/all-by-user

kaladay:
Perhaps we should also consider adding a "loading.." message just like is done when loading the submission?

This way there is feed back if there ever ends up being a performance problem again once this performance problem is solved.

jcreel:
We suspect this is caused by a database with two top-level organizations.

kaladay:
Investigation the SQL shows that there is a major problem with the Vocabulary Words as well.

There are 14 rows in `controlled_vocabulary` and there are 18395 rows in `vocabulary_word`.
For some reason, the code is going through seemingly every vocabulary word running an SQL select on each (I didn't bother to count).

Switching from EAGER to LAZY changed the performance on the page in question from ~22 seconds too ~600ms. But when this is done, then the loading a single submission takes forever (greater than 22 seconds, I just stopped waiting).

Attached is the SQL log dump taken from turning on all logging with the following log tags after refreshing the "Choose Your Organization" page.

(Github is giving problems with uploading the file, will try again after posting this comment.)

kaladay:
This is also happening on every start:
```
2023-06-07 17:06:41.280  INFO 34558 --- [           main] org.tdl.vireo.service.SystemDataLoader   : Loading default settings
2023-06-07 17:06:41.285  INFO 34558 --- [           main] org.tdl.vireo.service.SystemDataLoader   : Loading default Proquest language codes
2023-06-07 17:06:41.433  INFO 34558 --- [           main] org.tdl.vireo.service.SystemDataLoader   : Loading default Proquest degree codes
2023-06-07 17:06:41.463  INFO 34558 --- [           main] org.tdl.vireo.service.SystemDataLoader   : Loading default Proquest subject codes
2023-06-07 17:07:10.803  INFO 34558 --- [           main] org.tdl.vireo.service.SystemDataLoader   : Loading default Submission List Columns
[...]

Issue #1596, opened 2023-03-07
Performance


cstarcher:
Address performance issues in Submission, List and View

Edwin12335:
>

PR #1800, opened 2023-06-08
Issue 1756: Use Projection for specific Submission usage.

kaladay:
relates #1596
relates #1756

This is affecting the **Choose Your Organization** page.

Too much unnecessary data is being loaded off of the specific Submission end point `/all-by-user`. This end point is only used for the main page that then leads to the **Choose Your Organization** page.

This reduces the loaded data to only the minimum needed on the Student Submission List View. The observed difference in performance is ~800 milliseconds prior to this change and ~200 milliseconds after this change (for vocabulary words with ~17,000 contacts). Also note that Without the code from commit a2e872a6fb37a8bf1ede8cbb8210455a62ebff27, the prior performance is ~22 seconds (for vocabulary words with ~17,000 contacts).

### DIFF
--- a/src/main/java/org/tdl/vireo/controller/SubmissionController.java
+++ b/src/main/java/org/tdl/vireo/controller/SubmissionController.java
@@ -100,7 +100,7 @@ import org.tdl.vireo.service.SubmissionEmailService;
 import org.tdl.vireo.utility.OrcidUtility;
 import org.tdl.vireo.utility.PackagerUtility;
 import org.tdl.vireo.utility.TemplateUtility;
-import org.tdl.vireo.view.SimpleSubmissionView;
+import org.tdl.vireo.view.FieldValueSubmissionView;
 
 @RestController
 @RequestMapping("/submission")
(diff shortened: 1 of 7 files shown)


wwelling (commented):
This is an awesome example of using dynamic projection. However, I am not a fan of the additional named JPA method and interfaces to accomplish the same thing that is support using JsonView. 

I recommend, if not already attempted and if not a kludge to add annotation on the models, try adding an additional class to [`org.tdl.vireo.model.response.Views`](https://github.com/TexasDigitalLibrary/Vireo/blob/main/src/main/java/org/tdl/vireo/model/response/Views.java) and the appropriate JsonView annotation on the controller endpoint.

Alternatively, we begin building a suite of interfaces that solely interact as dynamic projection and cleanup the annotations in the models as this seems to be a cleaner separation of concerns in code. Implementing two different ways to do the same thing in code lacks consistency and clarity in how to read and maintain the code.

wwelling (approved):

wwelling:
As @kaladay pointed out, there is additional caveat between JsonView and interface dynamic projection from JPA. The JsonView only prevents LAZY fetched relations queries upon serialization when calling the getters. All EAGER and any manual calling getters will still perform unnecessary queries for this endpoint.

wwelling:
Upon further discussion, the appropriate solution would be to have a suite of interfaces that are implemented by the entities (multiple interfaces to decompose the entity and what is queried for). This would alleviate the disconnect between the interface signatures and the entity class method signatures.

cstarcher (approved):

Issue #1786, opened 2023-05-25
Public filters saved by another user do not work


cstarcher:
### Current behavior

When a user saves a public filter created by another user, the filter shows in the Saved Filters list but does not work.

### Expected behavior

A user should be able to save a public filter to their own profile as a separate private or public filter.

### Steps to reproduce

1. Apply a public filter in the list view
2. Save the filter as a new private or public filter
3. Apply the new filter

PR #1806, opened 2023-06-14
Issue 1795: Redesign filter broadcasting logic.

### LINKED ISSUE #1795: SPIKE: Active Filter List in Admin Submission View is not always refreshing.
When loading an active filter, the filter does not refresh.

Investigation has revealed that for the active filter, the UI is listening on the channel for the old filter ID. If the active filter changes in any way then a new broadcast is sent out using the new filter ID rather than the old filter ID. As a result, the UI does not refresh the Active Filter List (Now Filtering By).

Code prior to 4.2.1 is performing a `clone()` and broadcasting using the old filter ID with the cloned filter as the value associated with it.
The newer code is using the new filter ID when broadcasting rather than the old filter ID.
This is not a regression but is instead exposing an existing bug.

This problem likely exists in multiple places where the active filter is changed in some way and a broadcast is sent out to the channel.

The old behavior:
- https://github.com/TexasDigitalLibrary/Vireo/blob/8ff079ba606173e16782f418444ab4d65a497160/src/main/java/org/tdl/vireo/controller/SubmissionListController.java#L162

The updated behavior:
- https://github.com/TexasDigitalLibrary/Vireo/blob/5098dae99c65f84a955e6f958d55d8a4bf60fcfa/src/main/java/org/tdl/vireo/controller/SubmissionListController.java#L164

Example change in the new code that helps reveal the problem described above:
```java
    @RequestMapping(value = "/set-active-filter", method = POST)
    public ApiResponse setActiveFilter(@WeaverUser User user, @RequestBody NamedSearchFilterGroup namedSearchFilterGroup) {

        Optional<NamedSearchFilterGroup> desiredFilter = namedSearchFilterGroupRepo.findById(namedSearchFilterGroup.getId());

        if (desiredFilter.isEmpty()) {
            return new ApiResponse(ERROR, "Failed to find filter with ID " + namedSearchFilterGroup.getId() + ".");
        }

        NamedSearchFilterGroup activeFilter = user.getActiveFilter();

        user.setActiveFilter(desiredFilter.get());

        user = userRepo.update(user);

        if (activeFilter != null) {
            simpMessagingTemplate.convertAndSend("/channel/active-filters/" + activeFilter.getId(), new ApiResponse(SUCCESS, user.getActiveFilter()));
        }

        simpMessagingTemplate.convertAndSend("/channel/active-filters/" + user.getActiveFilter().getId(), new ApiResponse(SUCCESS, user.getActiveFilter()));

        return new ApiResponse(SUCCESS);
    }
```

Notice how this example is sending both the old and the new.

What likely needs to happen is to not change the back end but instead change the front end.
The UI should probably listen on the user model and watch for changes.
As an example, if there is a new listener channel, such as `/channel/user/[user id]/active-filter`, then this could be listened on.

An alternative solution could be to use the user ID rather than the filter ID in the channel.
The active filters are each associated with a single user and so using such a channel the user ID can consistently designate that the active filter for some user has been changed.

This would look like:
```java
simpMessagingTemplate.convertAndSend("/channel/active-filters/" + user.getId(), new ApiResponse(SUCCESS, user.getActiveFilter()));
```

Also note that the java code snippet far above is an example only.
Using it reveals that the first message is received but then follow up messages fail.
There is likely more to the problem than what is recorded here thus far.

kaladay:
resolves #1795 

This does not address the columns on the table.
This does not address the Filter By Display Filter boxes (which likely should follow the same or similar solution).

The weaver model listener operates on an individual basis.
Broadcasts received are on a path such as `/channel/active-filters/1`.
This only works so long as the active filter is filter id 1 (based on the above "such as" example).
Any time this changes, the listener could be listening on a different id than what is now the active filter as far as the database is concerned.
The saved filters have the same concern as the active filters.

The problem is that the model listener is on a per individual filter basis.
The entire set for both saved filters and active filters for a given user needs to be handled as a whole.

Redesign the logic to:
1. Disable the active filter model listener.
2. Utilize a new group-level (or set-level) channels: `/channel/active-filters/user/[user id]` and `/channel/saved-filters/user/[user id]` (where `[user id]` represents some valid user id).
3. Add a new FilterAction enumeration to communicate purpose.
4. Always return the Active Filter or Saved Filter.
5. Update active filters, saved filters, and table (rows) using the response received by the channel broadcast.
6. Remove now duplicate `query()` calls, preventing double-loading.
7. Remove call to non-existent function `activateFilter()` in filterOptions.html.
8. Remove `console.warning()` as that code is now required.

The listener is updating based on the response received.
That response may entail clearing the pagination and so an enumeration called `FilterAction` is provided.
When the listener begins to update, the current page could be past the max total pages.
A listener on a page other than the one triggering the broadcast has no way of knowing what is correct until it receives a response. Therefore, once a response is received and the page number is determined to be incorrect, then update the page number. This behavior is necessary and so the console warning is now removed.

### DIFF
--- a/src/main/java/org/tdl/vireo/controller/SubmissionListController.java
+++ b/src/main/java/org/tdl/vireo/controller/SubmissionListController.java
@@ -194,10 +194,17 @@ public class SubmissionListController {
             return new ApiResponse(ERROR, "Cannot delete filter, you do not own filter with ID " + filterGroup.get().getId() + ".");
         }
 
+        Boolean isPublic = filterGroup.get().getPublicFlag();
+        Long fgId = filterGroup.get().getId();
+
         namedSearchFilterGroupRepo.delete(filterGroup.get());
 
         simpMessagingTemplate.convertAndSend("/channel/active-filters/user/" + user.getId(), new ApiResponse(SUCCESS, FilterAction.REFRESH, user.getActiveFilter()));
         simpMessagingTemplate.convertAndSend("/channel/saved-filters/user/" + user.getId(), new ApiResponse(SUCCESS, FilterAction.REFRESH, user.getSavedFilters()));
+
+        if (isPublic == true) {
+            simpMessagingTemplate.convertAndSend("/channel/saved-filters/public", new ApiResponse(SUCCESS, FilterAction.REMOVE, fgId));
+        }
 
         return new ApiResponse(SUCCESS, user.getActiveFilter());
     }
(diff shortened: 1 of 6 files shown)


cstarcher (approved):

kaladay:
This appears to also solve #1786

PR #1812, opened 2023-06-21
Issue 1811: The full model needs to be defined when resetting so that the functions exist.

### LINKED ISSUE #1811: Cannot save changes when editing Deposit Locations.
When trying to edit a Deposit Location using the modal, then Save fails due to missing javascript function.

The console shows this error:
```
0000000005.js:15697 TypeError: $scope.modalData.save is not a function
```

Which happens here:
```js
        $scope.updateDepositLocation = function () {
            $scope.modalData.save();
        };
```

This is likely the same or similar problem that is affecting issue #1537.

kaladay:
resolves #1811

The methods like `save()` are from the `DepositLocation` model. The reset method is using a basic Javascript object.

Wrap the reset method data model initialization in a `DepositLocation` model.

### DIFF
--- a/src/main/webapp/app/controllers/settings/depositLocationRepoController.js
+++ b/src/main/webapp/app/controllers/settings/depositLocationRepoController.js
@@ -37,7 +37,7 @@ vireo.controller("DepositLocationRepoController", function ($controller, $scope,
                 $scope.modalData.refresh();
             }
 
-            $scope.modalData = {
+            $scope.modalData = new DepositLocation({
                 name: '',
                 depositorName: 'SWORDv1Depositor',
                 timeout: 240,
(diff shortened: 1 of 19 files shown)


src/main/webapp/app/controllers/settings/depositLocationRepoController.js
```diff
@@ -37,7 +37,7 @@ vireo.controller("DepositLocationRepoController", function ($controller, $scope,
                 $scope.modalData.refresh();
             }
 
-            $scope.modalData = {
+            $scope.modalData = new DepositLocation({
```
wwelling, line 40: Good catch. This is the intended design.

src/main/webapp/tests/mocks/models/mockDepositLocation.js
```diff
@@ -94,6 +94,17 @@ var mockDepositLocation = function($q) {
         return payloadPromise($q.defer(), testConnectionPayload);
     };
 
+    model.testDepositLocation = function() {
```
wwelling, line 97: Just to note, these methods are augmented methods and do not belonging to the model/prototype. They require the controller scope and hence violating object-oriented principles of loose coupling. For cleaner code, they should be moved to the model and add arguments (monadic or dyadic preferred) to avoid this. Therefore, the unit tests would be against the model for these and mocked for the controller.

kaladay: This is addressed here:
- https://github.com/TexasDigitalLibrary/Vireo/pull/1818

rmathew1011 (approved):

PR #1816, opened 2023-06-27
Issue 1815: Add "Archived" Submission status.

### LINKED ISSUE #1815: Vireo3 to Vireo4 ActionLog Migration may fail due to ARCHIVED Submission Status.
The Submission Status has evolved into a boolean column in the `submission_status` table in Vireo 4.

This is fine except for when migrating from Vireo 3 to Vireo 4.
There are Action Logs that have the `Archived` status.

It is unclear as to what this ActionLog should be moved to and so there is a need to still have an `Archived` Submission Status if only for the purposes of maintaining migrated data.

If Vireo 3 is not to be migrated then this likely should not be used and is not needed.

The System Data Loader can configure this but the values is also a hard-coded enumeration to deal with.

kaladay:
resolves #1815 

Without this several action logs may fail to migrate when migrating from Vireo 3 into Vireo 4.

The SQL to manually add this to the DB is:
```sql
insert into submission_status (id, is_active, is_archived, is_deletable, is_editable_by_reviewer, is_editable_by_student, is_publishable, name, submission_state) values (13, false, true, true, false, false, false, 'Archived', 13);
```

### DIFF
--- a/src/main/java/org/tdl/vireo/model/SubmissionState.java
+++ b/src/main/java/org/tdl/vireo/model/SubmissionState.java
@@ -13,7 +13,8 @@ public enum SubmissionState {
 	PUBLISHED(9), 
 	ON_HOLD(10),
 	WITHDRAWN(11),
-	CANCELED(12);
+  CANCELED(12),
+  ARCHIVED(13);
 	
 	private int value;
 
--- a/src/main/java/org/tdl/vireo/model/repo/impl/SubmissionRepoImpl.java
+++ b/src/main/java/org/tdl/vireo/model/repo/impl/SubmissionRepoImpl.java
@@ -244,6 +244,8 @@ public class SubmissionRepoImpl extends AbstractWeaverRepoImpl<Submission, Submi
             break;
         case WITHDRAWN:
             break;
+        case ARCHIVED:
+            break;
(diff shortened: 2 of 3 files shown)


kaladay:
Should not be needed.
Analysis of the data has revealed that most of the data is in the `PUBLISHED` status immediately before being switch into the `ARCHIVED` status or has a Deposit ID set on the Submission.

The migration can be changed to migrate `ARCHIVED` into `PUBLISHED` without needing to perform any code changes.

### Relevant SQL

Get count of Submissions with a Deposit ID:
```sql
select count(*) from submission where id in (select distinct submission_id from actionlog where submissionstate = 'Archived') and depositid is not null;
```

Get count of Submissions without a Deposit ID:
```sql
select count(*) from submission where id in (select distinct submission_id from actionlog where submissionstate = 'Archived') and depositid is null;
```

Get all of the Action Logs that have a Submission without a Deposit ID and has an `ARCHIVED` status:
```sql
select id, actiondate, submission_id, submissionstate from actionlog where submission_id in (select id from submission where id in (select distinct submission_id from actionlog where submissionstate = 'Archived') and depositid is null) group by submission_id, id order by submission_id asc, actiondate desc;
```

PR #1853, opened 2023-07-25
Issue 1842: Use JPA and include order by on vocabulary when ordering Controlled Vocabulary dictionary and revert input select.

kaladay:
Relates #1851

Use JPA instead of Hibernate to perform the OrderBy. The Hibernate uses SQL and JPA is the preferred approach.

The Degree information still needs sorting, add back orderBy for input select.

At least if most of the information is sorted already, then there will still be a performance boost of not having to move anything around in the case of the input select.

### DIFF
--- a/src/main/webapp/app/views/inputtype/input-select.html
+++ b/src/main/webapp/app/views/inputtype/input-select.html
@@ -9,7 +9,7 @@
     <li class="select-option"
       role="menuitem"
       ng-click="fieldValue.value = word.name; fieldProfileForm.$dirty = true; saveWithCV(fieldValue, word);"
-      ng-repeat="word in profile.controlledVocabulary.dictionary | filter:displayVocabularyWord">
+      ng-repeat="word in profile.controlledVocabulary.dictionary | filter:displayVocabularyWord | orderBy:'name'">
       <a href="#">{{word.name}}</a>
     </li>
   </ul>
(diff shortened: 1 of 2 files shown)


src/main/webapp/app/views/inputtype/input-select.html
```diff
@@ -9,7 +9,7 @@
     <li class="select-option"
       role="menuitem"
       ng-click="fieldValue.value = word.name; fieldProfileForm.$dirty = true; saveWithCV(fieldValue, word);"
-      ng-repeat="word in profile.controlledVocabulary.dictionary | filter:displayVocabularyWord">
+      ng-repeat="word in profile.controlledVocabulary.dictionary | filter:displayVocabularyWord | orderBy:'name'">
```
wwelling, line 12: Are there other locations where `orderBy:'name'` sorting removed recently?

kaladay: See the PR referenced in the comments above.
Linked here for convenience:
- https://github.com/TexasDigitalLibrary/Vireo/pull/1851

wwelling: Yes, thank you. That is the PR I glanced at. That PR changed multiple template loop expressions. Do any of those need to be reverted back? Or just the one?

kaladay: Just this one as this PR states.

rmathew1011 (approved):

PR #1854, opened 2023-07-26
Issue 1842: Submission create is taking longer by processing unneeded data.

### LINKED ISSUE #1842: Improve performance of Submission and Organization load impaired by large CV

kaladay:
Partially resolves #1842

The Submission create takes a long time because the back end is returning the full created Submission model. The front end then processing the entire Submission model. This takes a long time.

The processed model is added to the Submission repo but then a redirect is performed on the ID. All of that just created Submission model data is lost due to the redirect. Loading and processing that data is a waste of time.

Redesign the back end to instead return an ID on create rather than the full model. Instead, return just the newly created Submission ID. On redirect, the new Submission is fully loaded anyway and so only getting the ID is not a problem.

This saves ~2000ms in my local test case.

### DIFF
--- a/src/main/java/org/tdl/vireo/controller/SubmissionController.java
+++ b/src/main/java/org/tdl/vireo/controller/SubmissionController.java
@@ -240,7 +240,7 @@ public class SubmissionController {
     );
     actionLogRepo.createPublicLog(submission, user, "Submission created.");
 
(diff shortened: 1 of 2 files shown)


wwelling (approved):
I would double check if any other scopes make the same request and use the full submission response. This makes sense to me. 201 Created empty body or URL or identifier.

PR #1859, opened 2023-07-27
Issue 1858: Set wysiwyg parameter on all lockingtextarea fields.

### LINKED ISSUE #1858: Lock/Unlock Text Area Fields on Organization Management are not usable.
This has the same problem as the Look and Feel bug described here:
- #1776

When the unlock is clicked, there is no editable text field.

This is for all such fields on under Organization in the admin settings view.

kaladay:
Resolves #1858

This applies the same fix from PR #1792 (commit ef5004bd7b6a4ec5c37a159e9693893a722aaa6f).

The wysiwyg attribute needs to be specified now on the `lockingtextarea` field.

### DIFF
--- a/src/main/webapp/app/views/admin/settings/organization/workflowsteps/workflowStepContent.html
+++ b/src/main/webapp/app/views/admin/settings/organization/workflowsteps/workflowStepContent.html
@@ -1,7 +1,7 @@
 <div class="row clearfix" ng-if="step.overrideable || step.originatingOrganization === getSelectedOrganization().id">
 
-	<div class="col-md-4">
-		<validatedinput
+    <div class="col-md-4">
+        <validatedinput
             no-id="true"
             type="text"
             label="Workflow Step Name"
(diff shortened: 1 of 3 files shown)


wwelling (approved):
Hard to argue with being explicit with attributes.

However, the PR description assertion

>The wysiwyg attribute needs to be specified now on the lockingtextarea field.

is unconfirmed.

The [commit history](https://github.com/TexasDigitalLibrary/Vireo/commits/4.2.4_staging/src/main/webapp/app/directives/lockingTextAreaDirective.js) of the file is fun though.

wwelling:
Perhaps [ng-switch](https://docs.angularjs.org/api/ng/directive/ngSwitch) in the template can provide a default.

https://github.com/TexasDigitalLibrary/Vireo/blob/4.2.4_staging/src/main/webapp/app/views/directives/lockingTextArea.html

Has this been tried to not require `wysiyg` being specified?
 
```
                        <div ng-switch="wysiwyg">
                        <div ng-switch-when="true">
				<textarea 	ui-tinymce="tinymceOptions" 
							ng-model="$parent.scopeValue">
				</textarea>
			</div>
			<div ng-switch-default>
			          <textarea 	class="form-control locking-text-area-not-wysiwyg" 
							ng-model="$parent.scopeValue" 
							ng-keypress="$parent.nonWysiwygTyping($event)"
							ng-blur="$parent.nonWysiwygBlur()">
				</textarea>
			</div>
			</div>
```

Issue #1776, opened 2023-05-11
Look and feel text edit box not showing after clicking lock to edit


cstarcher:
After clicking the lock on the following elements in Look and Feel, the text edit box disappears, and the text cannot be edited.

Front Page Instructions (before button)
Front Page Instructions (after button)
Post Submission Instructions
Post Correction Instructions

![Screenshot 2023-05-11 at 10 18 19 AM](https://github.com/TexasDigitalLibrary/Vireo/assets/5767401/9cee7100-5278-4258-8ca1-187b783010ef)

PR #1862, opened 2023-07-28
Issue 1858: Fix WYSIWYG in lock unlock fields, reverting previous approach.

### LINKED ISSUE #1858: Lock/Unlock Text Area Fields on Organization Management are not usable.

kaladay:
resolves #1858 

Originally, the Locking Text Area fields were not working at some point in time.
This got fixed by disablying the wysiwyg attribute in this PR:
- https://github.com/TexasDigitalLibrary/Vireo/pull/1792

(For this issue: https://github.com/TexasDigitalLibrary/Vireo/issues/1776)

The lack of a `wysiwyg` attribute suggested that wysiwyg is supposed to be false.

That code was approved and merged in previous sprints.

The same bug was discovered a second time in different places and recorded here:
- https://github.com/TexasDigitalLibrary/Vireo/issues/1858

The solution chosen is to do the same exact approach as the previously approved code:
- https://github.com/TexasDigitalLibrary/Vireo/pull/1859

The problem is not the "=== 'wysiwyg'" and "!== 'wysiwyg'" checks.

Instead, the WYSIWYG data is not being loaded by the webpack.
This is essentially a regression from when the project switched to use webpack.

The "custom css" text area is left to have wysiwyg set to false.

### DIFF
--- a/.wvr/build-config.js
+++ b/.wvr/build-config.js
@@ -31,6 +31,10 @@ const config = {
       to: './node_modules/@wvr/core/app/views'
     },
     {
+      from: './node_modules/tinymce/skins',
+      to: './skins'
+    },
+    {
       from: './build/appConfig.js.template',
       to: './appConfig.js',
       transform(content) {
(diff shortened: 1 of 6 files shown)


jcreel (approved):
WYSIWYG tested and working.

PR #1873, opened 2023-11-22
sql query fix

### LINKED ISSUE #1872: SQL error when using certain queries
The file src/main/java/org/tdl/vireo/model/repo/impl/SubmissionRepoImpl.java is missing critical whitespace in StringBuilder assembled queries.  A space needs to be added after 'ORDER BY' in line 949 and after the 'AS' in lines 1089 and 1091.

cstarcher:
Add missing spaces for dates strings in SQL query.

Resolves #1872

### DIFF
--- a/src/main/java/org/tdl/vireo/model/repo/impl/SubmissionRepoImpl.java
+++ b/src/main/java/org/tdl/vireo/model/repo/impl/SubmissionRepoImpl.java
@@ -1091,13 +1091,13 @@ public class SubmissionRepoImpl extends AbstractWeaverRepoImpl<Submission, Submi
                 value.append("CAST(REPLACE(").append(table).append(".value, ' ', ' 1, ') AS DATE) AS");
             }
 
-            value.append(table).append("_date");
+            value.append(" " + table).append("_date");
 
             if (!sqlAliasBuilders.contains(value.toString())) {
                 sqlAliasBuilders.add(value.toString());
             }
 
-            sqlOrderBysBuilder.append(table).append("_date ").append(sort.name()).append(",");
+            sqlOrderBysBuilder.append(" " + table).append("_date ").append(sort.name()).append(",");
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/tdl/vireo/model/repo/impl/SubmissionRepoImpl.java
```diff
@@ -1091,13 +1091,13 @@ public void setColumnOrderingForMonthYearDateFormat(Sort sort, List<String> sqlA
                 value.append("CAST(REPLACE(").append(table).append(".value, ' ', ' 1, ') AS DATE) AS");
             }
 
-            value.append(table).append("_date");
+            value.append(" " + table).append("_date");
```
kaladay, line 1094: The design here is using string builder to avoid string concatenation.
I suggest instead changing it to add a space on lines 1089 and 1091 like this:
```
value.append("PARSEDATETIME(").append(table).append(".value, 'MMM yyyy') AS ");
```
and
```
value.append("CAST(REPLACE(").append(table).append(".value, ' ', ' 1, ') AS DATE) AS ");
```

Note that the the space is added after ` AS"`.

src/main/java/org/tdl/vireo/model/repo/impl/SubmissionRepoImpl.java
```diff
@@ -1091,13 +1091,13 @@ public void setColumnOrderingForMonthYearDateFormat(Sort sort, List<String> sqlA
...
 
             if (!sqlAliasBuilders.contains(value.toString())) {
                 sqlAliasBuilders.add(value.toString());
             }
 
-            sqlOrderBysBuilder.append(table).append("_date ").append(sort.name()).append(",");
+            sqlOrderBysBuilder.append(" " + table).append("_date ").append(sort.name()).append(",");
```
kaladay, line 1100: As stated in above comment, attempt to avoid literal string concatenation and instead use StringBuilder to add the space for consistency reasons.
```suggestion
            sqlOrderBysBuilder.append(" ").append(table).append("_date ").append(sort.name()).append(",");
```

kaladay (changes requested):

wwtamu (approved):
LGTM

PR #1887, opened 2023-12-29
[Issue 1883,1879,1878,1877,1876,1766,1682] 1878 continued - System data loader refactor, cleanup, and additional patches

### LINKED ISSUE #1682: After deleting a file, the screen is not scrollable.
This happens on the student submission view.

I have tested this across different browsers and found the problem to be consistent.

### Steps to reproduce:
1) Login and create a submission.
2) Upload a file to that submission.
3) Delete that file.
4) Attempt to scroll after the delete modal closes.

### LINKED ISSUE #1766: Choose Your Organization triptych not keyboard accessible
Using keyboard navigation does not allow selection of organizations in the triptych under "Choose Your Organization." This also prevents the screenreader from reading the triptych content.

The triptych should gain focus after the top-level institution by using the tab key. Focus should also move between triptych boxes using the tab or arrow keys. The organization should be selectable using the enter or spacebar key. Once a selection has been made, the tab key should move focus to the "Start Submission" button.

### LINKED ISSUE #1876: Deleted workflow steps not deleted from List View Customize Column
### Current behavior
When a user deletes an unused Field profile from a Workflow Step, the Field profile is not removed from the Customize columns list nor is the associated Metadata Key removed from the Workflow Management Metadata Keys.

### Expected behavior
An unused Field profile that has been deleted from all workflows should also be removed from the Customize columns list and the Workflow Management Metadata Keys.

### Steps to reproduce

1. Add a Field profile to a Workflow Step.
2. Remove Field profile from Workflow Step.
3. Open Customize columns modal in List view. The Field is still listed.
4. Open Metadata Keys under Workflow Management. The associated Metadata Key is still listed.

### LINKED ISSUE #1877: ORCiD not authenticating correctly
### Current behavior
When a user enters a valid ORCiD and the `Authenticate Orcid ID:` setting is set to `YES` in `Workflow Management` the ID is not being authenticated correctly. The user sees an error the value is not saved.

Screenshot 2023-12-20 at 3 27 46 PM

### Expected behavior
When a user enters a valid ORCiD and the `Authenticate Orcid ID:` setting is set to yes in `Workflow Management` the ID should authenticated correctly.

### Steps to reproduce

1. Set `Authenticate Orcid ID:` setting to `Yes` in `Workflow Management`.
2. Start a new submission.
3. Enter valid ORCiD into `ORCID:` field.
4. Tab or click out of field.

### LINKED ISSUE #1878: Rebooting resets email workflow rules
[*] Bug Report

Current Behavior:
On reboot, any custom email workflow rules disappear from Manage Email Workflow Rules. Two system workflow rules get added (but not enabled) to "submitted" status.

Desired Behavior:
Custom email workflow rules stay after reboot and default system email rules don't get added.

Relevant Information:
We're self hosted and use all custom email template rules.

System CVs also get populated with default values on reboot; we created all custom vocabs to work around this issue instead of editing existing vocabs (for example, "Major").

### LINKED ISSUE #1879: A fade covers the screen after deleting a thesis
[*] Bug Report

Current Behavior:
After deleting a submission (cancel submission > delete submission > select check to confirm), a fade covers the screen and prevents the admin from any further action. Refreshing the browser clears the fade.

Desired Behavior:
Display List after deleting a submission without the need to refresh. 

Relevant Information:
Tested on both Chrome and Firefox.

### LINKED ISSUE #1883: Give the submitter more information when file exceeds max size and upload fails
Feature enhancement

Current Behavior:
When a submitter uploads a file that exceeds the max size, they get an error that says only: "Upload Failed." They then contact us trying to understand why it failed, we have to spend time troubleshooting, and eventually we figure out it is because the file size is too big.

Desired Behavior:
Display a message with more information, like "Upload Failed. File exceeds max size [pull max size from application setting]." 
Would also be useful to give more information if they upload a file type that isn't accepted.

wwelling:
Resolves #1883 
Resolves #1879
Resolves #1878
Resolves #1877
Resolves #1876
Resolves #1766 
Resolves #1682


cstarcher:
I have reviewed the PR with the following results:

#1883 - Resolved
#1879 - Resolved
#1877 - Resolved
#1682 - Resolved

#1878 - I would like @smutniak to review this to confirm resolution.

#1876 - This fully resolves the issue as described. However, when deleting a field that is already being used as a `Customize Column` in `List` view, an error message is generated and the field is deleted from the `Workflow Step` field list but not from the `Customize Columns`. This creates a field value that exists but cannot be removed from the `Customize Columns`. The solution would be to validate that the field has not been assigned to the `List` view before deletion.

#1766 - This fully resolves the issue as described. However, when there are multiple workflows listed in the triptych, subsequent tabs moves to the last workflow before moving to the `Start Submission` button preventing the user from selecting any workflows except the last one.

smutniak (approved):
I've done some testing an review as best i could but the real test will be on a few sites incurring problems.  I'll merge into the staging branch and try a few deploys.  This PR superceds 1886 correct?

cstarcher:
I am receiving this error when I select the `Default columns` button in List view and in the `Customize Results Table` modal.

2024-01-18 13:45:37.930 ERROR 64188 --- [nio-9000-exec-3] o.t.v.c.SubmissionListController         : null
java.lang.UnsupportedOperationException: null
	at java.base/java.util.Collections$UnmodifiableCollection.clear(Collections.java:1079) ~[na:na]

PR #1893, opened 2024-01-22
Return new list from default submission list column service to avoid …

wwelling:
…mutation


cstarcher:
Resolves https://github.com/TexasDigitalLibrary/Vireo/pull/1887#issuecomment-1899108851

PR #1894, opened 2024-01-22
Prevent field profile delete when submission list column is in use

wwelling:


cstarcher:
Resolves concern over #1876 as discussed here: https://github.com/TexasDigitalLibrary/Vireo/pull/1887#issuecomment-1881139437

No further validation is needed to prevent deletion.

PR #1910, opened 2024-03-07
dspace fix

smutniak:
DSpaceMETS dc.date.issued date rendered as iso8601 format

### DIFF
--- a/src/main/java/org/tdl/vireo/utility/SubmissionHelperUtility.java
+++ b/src/main/java/org/tdl/vireo/utility/SubmissionHelperUtility.java
@@ -301,13 +301,12 @@ public class SubmissionHelperUtility {
         return year;
     }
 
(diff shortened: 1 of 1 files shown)


cstarcher (changes requested):
The dc.date.issued date in the DSpaceMETS export is still formatted as Month YYYY.

This PR is changing the following:

DSpaceMETS (mets.xml)

dc.date.created from YYYY-MM to YYYY-MM-DD
dc.date.submitted from Month YYYY to YYYY-MM-DD

DSpaceSimple (dublin_core.xml)

dc.date.created from Month YYYY to YYYY-MM-DD
dc.date.issued from YYYY-MM to YYYY-MM-DD.

The solution for #1911 should only change the dc.date.issued date format in the DSpaceMETS mets.xml file to YYYY-MM.

PR #1929, opened 2024-05-06
Cli submission generate improvements.

kaladay:
This is a collection of several CLI improvements, mostly focused on Submission generation.

### Relocate `generate` CLI code outside of transaction and improve error handling.
The relocation allows for faster transactional operation. With the loop outside of the transaction, then the transaction can be committed for each pass. This then allows for reduced memory requirements for large transactions (such as when generate a huge amount of submissions).

The exception handling is improved by catching the exceptions at the CLI level such that CLI related exceptions can be reported without crashing the entire server. Now, if a CLI fails due to some error, then the server does not exit.

### Improve CLI generation, fixing embargoes to be actual embargoes and several others.
The embargoes are now generated from the actual list of embargoes. If there are no embargoes then it falls back to a randomly generated string.

Handle other similar cases that are easy to resolve at the same time. This makes the generated data more accurate and easier to test. Especially when using them for filtering.

### Switch to `localhost.localdomain` for example address.
This is done to avoid any potential accidents if e-mails are sent.

### Improve performance of CLI generate.
Doing a `findAll()` and then counting the length of the array is highly inefficient.
Using `findAll()` slowing down the generation of large data sets for testing.
Change the code to an SQL count via the JPA `count()` method.

### DIFF
--- a/src/main/java/org/tdl/vireo/cli/Cli.java
+++ b/src/main/java/org/tdl/vireo/cli/Cli.java
@@ -2,12 +2,14 @@ package org.tdl.vireo.cli;
 
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Random;
 import java.util.Scanner;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.CommandLineRunner;
 import org.springframework.core.Ordered;
 import org.springframework.core.annotation.Order;
 import org.springframework.stereotype.Component;
+import org.tdl.vireo.model.User;
 import org.tdl.vireo.service.CliService;
 
 /**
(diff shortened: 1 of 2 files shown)


smutniak (approved):
I cannot test this because i no longer have an environment on my local computer to test spring-boot  - i test branches using an ansible built remote VM.    The code looks fairly straightforward so i'll approve and indirectly test this when I test 1936.

kaladay:
For reference, if using maven spring boot, the commands I use to test this are:
```
mvn clean spring-boot:run -Dspring-boot.run.arguments="console" -Dprofile="development"
expansive
generate 10
```

smutniak:
it worked on my remote server, i forgot some of the changes i did made '-Dprofile="development"' necessary

PR #1932, opened 2024-05-07
Issue 1931: Pagination not working properly for combined filters list.

### LINKED ISSUE #1931: Pagination not working properly for combined filters list (embargo and graduation semester)
A site admin reports:

When I was filtering by Semester and Default Embargo, and set to see 100 files per page, the search results at the bottom said "100 matching records displayed on 1 pages." But I knew there should be more, and there was no "next page" option that I could see, so I set my view to 200 files per page. Then the results said "200 matching records displayed on 1 pages." Still not enough, still no next page option, so I set to 400 per page, and got the number 205, which might indeed be correct. Is there no "next page" option at the moment when filtering?

I (frank) saw this on another site.  I filtered by 'default embargos' and then added 'graduation semester' and filtered by one instance of each.  There are 20 results.  When I select '10' as the number of records to display per page I see the message " 10 matching records displayed on 1 pages (10 per page). " and no list of pages (which is an error).  Clicking on 20 items per page I'll see 20 items on one page and a message of " 20 matching records displayed on 1 pages (20 per page). " and no list of pages (as it should be).  Clicking on 40 items per page I see " 20 matching records displayed on 1 pages (40 per page). " which is as it should be.

kaladay:
resolves #1931

This is a regression from the aggressive performance optimizations such as commit b2c538c5c9c7d99a18c9f8c2c47b8cc98c01a260.

The case that is not correctly handled is where there are multiple separate field values being selected and filtered upon.

Instead of fully removing the optimization, this uses the optimization when there are zero or one field_values being filtered on. Then when there are two or more field_values then omit the optimization.

### DIFF
--- a/src/main/java/org/tdl/vireo/model/repo/impl/SubmissionRepoImpl.java
+++ b/src/main/java/org/tdl/vireo/model/repo/impl/SubmissionRepoImpl.java
@@ -449,6 +449,7 @@ public class SubmissionRepoImpl extends AbstractWeaverRepoImpl<Submission, Submi
         sqlAliasBuilders.add("s.id");
 
         int n = 0;
+        int totalFieldValueConditions = 0;
 
         for (SubmissionListColumn submissionListColumn : allSubmissionListColumns) {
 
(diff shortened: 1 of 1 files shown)


smutniak (approved):

smutniak (approved):
still working

PR #1934, opened 2024-05-08
Unable to re-arrange the submission workflow in Manage Workflow.

kaladay:
The organization is being changed but the dirty flag is not being set. Manually set the dirty flag before sending.

The `OrganizationRepo` listeners are either not doing anything or they are not checking for when status is a failure.
The `ApiResponseActions.READ` and `ApiResponseActions.BROADCAST` never do anything because the `resetManageOrganization()` only does something when the organization is provided.
No organization is being provided.

The `OrganizationRepo.listen(ApiResponseActions.READ, function (res) {` now makes the function do something rather than being a no-op functon. The organization, if one is selected, is now reset with the selected organization.

The `BROADCAST` listener is removed.

### DIFF
--- a/src/main/webapp/app/controllers/settings/organizationManagementController.js
+++ b/src/main/webapp/app/controllers/settings/organizationManagementController.js
@@ -131,6 +131,13 @@ vireo.controller("OrganizationManagementController", function ($controller, $loc
                 $scope.resetWorkflowSteps();
             }
         });
+
+        OrganizationRepo.listen(ApiResponseActions.READ, function (res) {
+            var resObj = !!res && !!res.body ? angular.fromJson(res.body) : {};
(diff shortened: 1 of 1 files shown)


kaladay:
> i can reorganize items under specific steps of manage workflow though i see that the ordinals of the items do not updated until i actually refresh the page. is this expected?
> 
> I don't see other regressions but i won't be able to test it comprehensively like it will get on deployed sites though given the limited changes to the code i don't expect much regression. I do hope for unexpected bug fixes though.

We have this problem we (TAMU) migrated the data with "Hierarchical Organization Management" enabled.
I'll send you a video.

smutniak:
i rebuilt and tested following your vidoe (misunderstood what i was testing before) and now see that it works.

smutniak (approved):

PR #1935, opened 2024-05-08
Issue 1930: Add custom filter for embargo type by list.

### LINKED ISSUE #1930: Embargo Filters require unassisted text entry instead of a list like vireo3, or typeahead
As reported by a site admin:

When I customize filter by Default Embargo type, it appears I have to manually input every embargo type. On Vireo 3 I could select each embargo type from a list, but here I have to type "Journal Hold," "Other Hold," "Patent Hold," and "Creative Writing." Two of those holds are entered twice with the same name on the list of possible embargoes students can select, so I'm not sure whether my typing them is pulling all of them up. It seems preferable to be able to choose each type from a list when filtering.

I (frank) tested this on a different hosted site and found the same lack of dropdown list.

kaladay:
resolves #1930

There is existing (dead) code for this already.
Re-vitalize the (dead) code, clean it up, and get it working.

The `embargoTypes.name` SQL in the back-end operates differently in that it is not involved in part of a column name and is exclusively for filtering. This changes the nature of the query and I opted to go with sub-queries for selecting the IDs because I can design it this way and avoid `LEFT JOIN` (or `INNER JOIN`). This has not seen extensive performance testing but this new SQL query conditions should be run through some performance testing.

I went with `Embargo Type` for the new custom filter because the existing code used that name.

A custom filter `uniqueEmbargoType` is created because the `unique`/`uniq` angularjs is not available. Using `unique` or `uniq` yields an error like this:
```
Unknown provider: uniqueFilterProvider <- uniqueFilter
```

Using the custom filter (`uniqueEmbargoType`) avoids having to pass the additional ` | filter:{isActive: true}` and similar. This can be done and is done within the same filter.

This adds a unit test for the filter (AngularJS context of filter) for the Embargo Type filter (Vireo table list context of filter).

The addition of the `Embargo Type` causes a test in the `SystemDataLoaderTest` class to fail due to a hard-coded value representing all of the default filters. This has been updated to include the new filter in the total.

The UI has an existing filter view, and that filter view is now slightly tweaked to address problems in the user experience. I opted to make its design more closely match that of the Status filter (which goes in line with the issue request using Status filter as an example).

### DIFF
--- a/src/main/java/org/tdl/vireo/model/repo/impl/SubmissionRepoImpl.java
+++ b/src/main/java/org/tdl/vireo/model/repo/impl/SubmissionRepoImpl.java
@@ -740,41 +740,29 @@ public class SubmissionRepoImpl extends AbstractWeaverRepoImpl<Submission, Submi
                     break;
 
                 case "embargoTypes.name":
-                    sqlBuilder = new StringBuilder()
-                        .append("\nLEFT JOIN")
-                        .append("\n   (SELECT e.id, e.name, semt.submission_id")
-                        .append("\n   FROM embargo e")
-                        .append("\n   LEFT JOIN submission_embargo_types semt")
-                        .append("\n   ON semt.embargo_types_id=e.id) embs")
-                        .append("\n   ON embs.submission_id=s.id");
-
-                    sqlJoinsBuilder.append(sqlBuilder);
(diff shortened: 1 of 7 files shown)


smutniak (approved):
it appears neutral as it didn't break anything but i won't really be able to test until PR 1936 so i'll approve.

PR #1933, opened 2024-05-08
Issue 1875: Hide search box filter.

kaladay:
Resolves #1875

The Search Box filter was previously disabled/hidden but the filter was overlooked.

This helps finish the process by hiding the search box from the filter.

PR #1936, opened 2024-05-08
Add custom filter for submission type by list. 

kaladay:
This depends on and effectively includes the following PRs:
- #1929
- #1933
- #1935

This borrows heavily on the work resolved for Issue 54 (see commit cbe080bcdc4477bc1f5f8ceeff27abbf0bb94528).

Most of the field values Controller end points are under the Submission Controller rather than its own Controller. This introduces complications because the UI is using "FieldValue" and mapping it to a Submission Controller. Getting the list of FieldValues, independent of a Submission, does not belong in the Submission Controller. There are a lot of work-arounds or other unusual design changes here to accommodate this situation. A new `FieldValueRepo` is added on the UI.

This adds new functionality for selecting all Field Values associated with a given Field Predicate. This is used to select the specific Field Predicate that is hard-coded as `submission_type`.

It turns out the escape string usage in `cbe080bcdc4477bc1f5f8ceeff27abbf0bb94528` is not entirely correct. The string does need to be escaped, but not the `%` and not the `_`. Only the slashes.
Change the `escapeString()` function to handle this use case.

The **Submission Type** column already exists.
Create a new one called `Submission Type (List)`.
This exposed a problem in the design where the names of the columns are used to generate the array key names and HTML attribute names. The use of `(` and `)` is not well supported there. Break out the sanitizer used in several places into a single function call and have that function call also handle `(` and `)`. Other cases are not handled and I am leaving those to be handled on an as-needed basis. To avoid conflicts with the existing **Submission Type**, the new **Submission Type (List)** is suffixed with `List` in several places.

The `submissionListColumns` variable in the `processUpdate` of the `SubmissionListController` is being used for both the **Manage Filter Columns** and the **Submission List Columns**. This is a problem where the new `Submission Type (List)` is only a Filter and is not a Column to select. The code had to be re-worked to address this.
I chose the quick and simple approach of making another variable. There may be better ways but this seems good enough at this time.

A recent change introduced setting `$dirty` on `$scope.getSelectedOrganization()` but that did not handle the case where `$scope.getSelectedOrganization()` is undefined or NULL. This broke some unit tests but is now fixed.

Update the CLI tests to generate Submission Types that are not always different. The Submission Type is created outside of the transaction loop to ensure only a small set is created. If there already exists Submission Types then no new ones are created and the existing ones are re-used. This allows for building generated test data on a system with existing data already populated.

The following SQL may be needed to updating existing repositories when not using the `ddl-auto` to update.
```sql
INSERT INTO submission_list_column (id,title,input_type_id) VALUES (DEFAULT,'Submission Type (List)', 1);
INSERT INTO submission_list_column_value_path (submission_list_column_id,value_path,value_path_order) VALUES (currval('submission_list_column_id_seq'),'submissionTypes',0)
INSERT INTO submission_list_column_value_path (submission_list_column_id,value_path,value_path_order) VALUES (currval('submission_list_column_id_seq'),'name',1)
```

### DIFF
--- a/src/main/java/org/tdl/vireo/cli/Cli.java
+++ b/src/main/java/org/tdl/vireo/cli/Cli.java
@@ -129,13 +129,15 @@ public class Cli implements CommandLineRunner {
                             Random random = new Random();
                             long idOffset = cliService.countUsers();
                             User helpfulHarry = null;
+                            boolean hasSubmissionTypes = false;
 
                             if (expansive) {
                                 helpfulHarry = cliService.createHelpfulHarry(idOffset++, CliService.EMAIL_DATE);
+                                hasSubmissionTypes = cliService.hasSubmissionTypes();
                             }
 
                             for (long i = itemsGenerated; i < num1 + itemsGenerated; i++) {
-                                cliService.operateGenerate(expansive, num2, random, idOffset, helpfulHarry, i);
+                                cliService.operateGenerate(expansive, num2, random, idOffset, helpfulHarry, hasSubmissionTypes, i);
                                 System.out.print("\r" + (i - itemsGenerated) + " of " + num1 + " generated...");
                             }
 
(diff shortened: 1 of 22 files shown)


cstarcher (approved):
This resolves the related issues.

PR #1946, opened 2024-06-22
added link functionality to tinymce

### LINKED ISSUE #1912: More advanced WYSIWYG editing options for Look & Feel settings
**Is your feature request related to a problem? Please describe.**
In Vireo 3, it was possible to use HTML source code in the "Front Page Instructions" (and other instruction sections). Vireo 4 now provides a WYSIWYG editor to facilitate that. However, the editor is very limited; it only allows the ability to apply paragraph, headings, and preformatted text, as well as bold, Italics, and underline formatting. There is no way to include more advanced HTML coding (like text colors, hyperlinks, special CSS styles) to text that is entered into the section.

**Describe the solution you'd like**
Ideally, I would like the option to define -- either by editing the HTML source code directly or using additional editor options -- for selected text, if necessary. This problem came up when I wanted to add a special note to the Front Page about a scheduled system upgrade. In Vireo 3, I could add this kind of note in a box with a specific background color, so it would be nice to replicate that in Vireo 4. I have written up some CSS code that could be used, but there is no way to select that from the WYSIWYG editor.

**Describe alternatives you've considered**
An alternative option would be to include a special section in the Look & Feel that is specific for system alerts, which could be toggled on/off as needed. This section could provide the ability to set the background and text colors, with the placement set by default it if is toggled on.

cstarcher:
Resolves #1912

### DIFF
--- a/src/main/webapp/app/directives/lockingTextAreaDirective.js
+++ b/src/main/webapp/app/directives/lockingTextAreaDirective.js
@@ -56,9 +56,9 @@ vireo.directive("lockingtextarea", function ($timeout) {
                         save();
                     });
                 },
-                toolbar1: "formatselect bold italic underline | bullist numlist undo redo",
+                toolbar1: "formatselect bold italic underline | bullist numlist undo redo | link unlink",
                 theme: "silver",
-                plugins: "autoresize",
+                plugins: "autoresize link autolink",
                 menubar: false,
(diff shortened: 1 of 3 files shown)


smutniak (approved):

smutniak (approved):
looks great!

Issue #1944, opened 2024-06-12
Replacement of special characters in filenames


smutniak:
A common problem which requires manual intervention are files with certain characters in the filename.  This problem is usually noticed during an export where, in vireo4's attempt to zip up the files associated with a submission, it is unable to find a specific file as denoted in field_value.value

A recent example was a file which had ':' in the name (e.g. "stuff-i-reviewed_on4:29:2024.pdf").  The file needed to be renamed by replacing ':' with '_' and updating the field_value before vireo could find it.  

Upon ingest a filename could be renamed or the list of characters to be escaped could be expanded.
 
Related to the more specific #1904 and #1909

PR #1984, opened 2025-01-23
Fix error handling for batch exports

jsavell:
When an error occurrs while building the export, the error response gets written directly to the output stream, resulting in a corrupted export file and no indication from the UI that there was an error.

This PR restructures the backend export process to only write the export to the output stream when there are no errors and return an appropriate status code and apiResponse when there are error(s). 

The front end has also been updated to only deliver the response as a file download if it's error free.

With these changes, the error response is now able to be caught and displayed by the standard alert handling in the UI.

### DIFF
--- a/src/main/java/org/tdl/vireo/controller/SubmissionController.java
+++ b/src/main/java/org/tdl/vireo/controller/SubmissionController.java
@@ -41,6 +41,7 @@ import org.springframework.data.domain.PageRequest;
 import org.springframework.data.domain.Pageable;
 import org.springframework.data.domain.Sort.Direction;
 import org.springframework.data.web.PageableDefault;
+import org.springframework.http.HttpStatus;
 import org.springframework.messaging.simp.SimpMessagingTemplate;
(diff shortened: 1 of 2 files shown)


cstarcher:
On review, check how this relates to #1944

cstarcher:
This fixes the problem of corrupted export files. The problem that remains is that there is no useful information provided for the user concerning the reason for the error. The export simply fails and the user sees only the attached error.
Screenshot 2025-03-11 at 8 01 51 PM

jsavell:
Yes, the Weaver UI library is unfortunately discarding the original error message provided by the Vireo back end as part of its FileService functionality.

So, the improvement of the error being detected by the UI and the modal being properly reset is the best currently available Vireo only solution.

A fix for Weaver UI is in progress, though, and should be included in the next release: https://github.com/TAMULib/Weaver-UI-Core/pull/250

At that point, updating Vireo to use the next Weaver UI release should be the only additional Vireo change needed to have the error message properly displayed to the user.

cstarcher (approved):

PR #1994, opened 2025-03-18
Account for ActionLog object type in ExcelPackager

jameswsullivan:
### Issue:

When exporting a list to Excel using `BATCH OPERATIONS` -> `Download Export` -> `Export Format` -> `Excel`, the `Last Event/Last Action` column in the exported Excel file would not show the actual text from the Action Log as seen in the UI, but shown as a reference to the object/instance instead, such as `org.tdl.vireo.model.ActionLog@1234ab`. The `ExcelPackager` didn't account for the `ActionLog` object type in the series of `else if` s.

### Solution:

Add a condition check for `ActionLog` type and retrieve the action log entry text.


cstarcher:
Resolves #1995

Issue #1995, opened 2025-03-18
Last Event/Last Action column not exporting correctly for Excel export


cstarcher:
**Describe the bug**
When exporting a list to Excel using BATCH OPERATIONS -> Download Export -> Export Format -> Excel, the Last Event/Last Action column in the exported Excel file would not show the actual text from the Action Log as seen in the UI, but shown as a reference to the object/instance instead, such as org.tdl.vireo.model.ActionLog@1234ab. The ExcelPackager didn't account for the ActionLog object type in the series of else if s.

Solution:
Add a condition check for ActionLog type and retrieve the action log entry text.

PR #1996, opened 2025-03-18
Issue #1995: Account for ActionLog object type in ExcelPackager

jameswsullivan:
### Issue:

When a list is exported to Excel via `BATCH OPERATIONS` -> `Download Export` -> `Export Format` -> `Excel` , the `Last Event/Last Action` column in the exported Excel file is shown as a reference to the instance/object, such as `org.tdl.vireo.model.ActionLog@1234ab` rather than the actual last event text as seen in the UI. The `ExcelPackager` didn't account for the `ActionLog` object type in its series of `else if` s.

### Solution:

Add a `if` check in the Excel packager for the ActionLog object type and retrieve the action log entry text.


cstarcher:
Resolves #1995

wwtamu (approved):

Issue #1943, opened 2024-05-31
Action Log needs to be included as CSV file with METS or Simple export


jcross2562:
**Describe the bug**
In Vireo 3, the entries in the Action Log (i.e., the full history of a submission item) were included in METS and Simple exports as a CSV file, but this step is not included in Vireo 4.

**To Reproduce**
1. Go to the Admin List view.
2. Set a filter for a small subset of records.
3. Scroll down to Batch Operations and select Download Export.
4. Select the DSpaceMETS option and click through to download the file.
5. Open the zipped file and select one of the item folders.
6. Verify in Vireo 4 that that only files that display are the XML files and any files uploaded by the student. (In Vireo 3, an "action-log" CSV file would also be included.)

**Expected behavior**
The history of the submission should be included as a CSV file to document the history of the submission, which can and should be retained for preservation purposes.

**Additional context**
Since Vireo is not intended for preservation or to be an active repository of files, there must be a way to export all of the information from each submission record that can be saved for posterity elsewhere. And exporting the action log is one certain way to ensure that the item's full history is captured as part of that process. Being able to capture this along with all of the other relevant files for each submission will ensure that published items can be effectively and systematically removed from Vireo after all other steps have been completed.

ztajoliDev:
Hello, but how do you expect DSpace to handle this action_log.csv file?
Do you expect this file to be uploaded along with the others?
And should it be possible to download it?
Are there any examples of people who have used this feature?

PR #2021, opened 2025-03-31
Sprint 8 staging 1943 actionlog

smutniak:
action_log.csv output added to DSpaceSimple and DSpaceMETS exports to solve #1943


wwtamu (changes requested):
Can we move the CSV file creation to a utility method?

Something like:

Csv.java
```
public static File fromActionLog(Set <ActionLog> actionLog) {
...
}
```

wwtamu (approved):

PR #2026, opened 2025-04-01
Add CSV utility to create CSV from action log

wwtamu:
Move CSV creation to CSV utility. Use Apache commons-csv. Add unit tests.

### DIFF
--- a/src/main/java/org/tdl/vireo/model/packager/DSpaceMetsPackager.java
+++ b/src/main/java/org/tdl/vireo/model/packager/DSpaceMetsPackager.java
@@ -1,5 +1,7 @@
 package org.tdl.vireo.model.packager;
 
+import javax.persistence.Entity;
+
 import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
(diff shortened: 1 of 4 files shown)


smutniak (commented):
Should User Name and Action Entry not be encapsulated in double quotes as there is no guarantee that either of them don't contain a ',' ?

wwtamu:
> Should User Name and Action Entry not be encapsulated in double quotes as there is no guarantee that either of them don't contain a ',' ?

https://commons.apache.org/proper/commons-csv/apidocs/org/apache/commons/csv/CSVPrinter.html#printRecord(java.lang.Object...)

> The values will be quoted if needed.

If we require quoting all values.

https://commons.apache.org/proper/commons-csv/apidocs/org/apache/commons/csv/CSVFormat.html#withQuoteMode(org.apache.commons.csv.QuoteMode).

https://commons.apache.org/proper/commons-csv/apidocs/org/apache/commons/csv/QuoteMode.html

smutniak (approved):

Issue #1562, opened 2022-04-12
Who Has Access and Add Member modal not updating without page refresh


cstarcher:
Relates to #1549
Relates to #1561

cstarcher:
Who Has Access

Two issues:

The modal is not updating when adding a user. I would expect that the selected user would be removed from the list after clicking the add button. The user is, however, added to the list.

The list does not update when assigning roles for STUDENT and ANONYMOUS. I would expect the user to be removed after selecting one of these roles. It only updates after I refresh the page.

cstarcher:
Reproduced in version 4.2.5.

PR #2040, opened 2025-04-08
Issue #1562: Update Who Has Access and Add Member response handling

wwtamu:
- remove broadcast all users on save user
- reload user tables on add member success
- add alert directive in modal to handle error and invalid responses


qtamu (approved):

kaladay (approved):
> remove broadcast all users on save user
Especially thank you for getting rid of that.

PR #2041, opened 2025-04-09
Issue #2032: Consolidate file upload get pattern and always return .pdf for `_doctype_primary`

wwtamu:
With this PR, the behavior of file upload will be as follows.

- if field profile predicate equals `_docype_primary` return pattern `.pdf`
- else if field profile has a controlled vocabulary, return constructed pattern from it
- else return `*`

In order to restrict file extensions for file upload of a new field profile that does not use `_docype_primary` field predicate, you can create a controlled vocabulary and assign it to the field profile for INPUT_TYPE.

### DIFF
--- a/src/main/webapp/app/controllers/submission/adminSubmissionViewController.js
+++ b/src/main/webapp/app/controllers/submission/adminSubmissionViewController.js
@@ -382,12 +382,10 @@ vireo.controller("AdminSubmissionViewController", function ($anchorScroll, $cont
         };
 
         $scope.getPattern = function (doctype) {
-            var pattern = "*";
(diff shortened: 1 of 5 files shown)


smutniak (approved):
I tested changing the primary document as admin and my upload window was restricted to pdf only.  Creating a submission in the workflow I saw my upload was also restricted to pdf for a new submission.

as admin i created a CV with 2 entries: 'txt' and 'png' and set 'supplemental files' to that CV restriction.  I created a new submission and those where the only 2 options i saw in my upload window.  If that is all there is to test it looks good to me.

Issue #2045, opened 2025-04-14
Check that all status changes trigger email workflow rules


cstarcher:
**Describe the bug**
Not all status changes are triggering related email workflow rules.

**Expected behavior**
There should be a check of email workflow rules after each status changes on a submission.

wwtamu:
Newly created submission is not sending for email workflow on status "In Progress".

https://github.com/TexasDigitalLibrary/Vireo/blob/main/src/main/java/org/tdl/vireo/controller/SubmissionController.java#L242

Batch publish is not checking email workflow rules whether to send after update status.

https://github.com/TexasDigitalLibrary/Vireo/blob/main/src/main/java/org/tdl/vireo/controller/SubmissionController.java#L810

Needs correction endpoint is not checking email workflow rules.

https://github.com/TexasDigitalLibrary/Vireo/blob/main/src/main/java/org/tdl/vireo/controller/SubmissionController.java#L903

wwtamu:
Publish endpoint is sending a duplicate email on error.

https://github.com/TexasDigitalLibrary/Vireo/blob/main/src/main/java/org/tdl/vireo/controller/SubmissionController.java#L511

PR #2046, opened 2025-04-15
Issue #2045: Check email workflow rules on submission status change to ensure they are respected

wwtamu:
It was found that not all persisted change of submission status was triggering email workflow rules. This is only a patch and does not guarantee all cases without a consistent and atomic approach is taken. There may be some missed by unexpected code paths and whether new code will not reintroduce this bug.


qtamu (approved):

wwtamu:
This PR may introduce performance concerns with batch publish as it will synchronously send emails for each submission in the batch as per any configured email workflow rules for transitioning to status published.

Issue #2048, opened 2025-04-17
Batch export issues


tamu-sad-iii:
- [x] unchecked null pointers when first and last name an empty string
- [x] large exports cause out of memory without streaming directly to response
- [x] duplicate zip entries with individuals with same first and last name and/or multiple identical license files
- [x] phone number parsing logs stack trace without debug enabled
- [ ] field values as asset paths are not all compatible with Windows and possibly Linux when containing `:` or `|` or ending in ` `
- [x] submissions without primary doc field value throw null pointer exception
- [x] close streams to avoid memory leak from exception
- [x] prevent large exports causing out of memory errors

wwtamu:
```
org.springframework.http.converter.HttpMessageNotWritableException: No converter for [class edu.tamu.weaver.response.ApiResponse] with preset Content-Type 'application/zip'
        at org.springframework.web.servlet.mvc.method.annotation.AbstractMessageConverterMethodProcessor.writeWithMessageConverters(AbstractMessageConverterMethodProcessor.java:312) ~[spring-webmvc-5.3.24.jar:5.3.24]
        at org.springframework.web.servlet.mvc.method.annotation.RequestResponseBodyMethodProcessor.handleReturnValue(RequestResponseBodyMethodProcessor.java:183) ~[spring-webmvc-5.3.24.jar:5.3.24]
```

```
java.util.zip.ZipException: duplicate entry: Lane_Matthew_1744838942669_permission/LICENSE.txt
        at java.base/java.util.zip.ZipOutputStream.putNextEntry(ZipOutputStream.java:244) ~[na:na]
        at org.tdl.vireo.controller.SubmissionController.processBatchExport(SubmissionController.java:700) ~[classes/:na]
```

```
java.lang.NullPointerException: Cannot invoke "org.tdl.vireo.model.FieldValue.getValue()" because "primaryDoc" is null
        at org.tdl.vireo.controller.SubmissionController.processBatchExport(SubmissionController.java:707) ~[classes/:na]
        at org.tdl.vireo.controller.SubmissionController.batchExport(SubmissionController.java:563) ~[classes/:na]
        at org.tdl.vireo.controller.SubmissionController$$FastClassBySpringCGLIB$$7dc1759.invoke(<generated>) ~[classes/:na]
        at org.springframework.cglib.proxy.MethodProxy.invoke(MethodProxy.java:218) ~[spring-core-5.3.24.jar:5.3.24]
```

```
[...]

PR #2054, opened 2025-04-22
Issue #2048: Batch export patches

wwtamu:
- [x] unchecked null pointers when first and last name an empty string
- [x] duplicate zip entries with individuals with same first and last name and/or multiple identical license files
- [x] phone number parsing logs stack trace without debug enabled
- [x] submissions without primary doc field value throw null pointer exception
- [x] close streams to avoid memory leak from exception thrown while building export
   > Solution required ensuring all streams are either within a try-with-resource or closed properly in a finally block.
- [x] prevent large exports causing out of memory errors
   > Solution required changes to storing the export directly to a temporary file and when complete streaming it to the response output stream.

The following issues are not addressed in this PR.

1. field values as asset paths are not all compatible with Windows and possibly Linux when containing `:` or `|` or ending in ` `

This issue must be addressed at the database and file system level. The files uploaded should be compatible with the file system running the application. To correct both, any files with these characters have to be renamed and the associated submission field profile value must be updated to match.

2. long running downloads causing timeouts

A solution to this could be to return the request to the client requesting export while asynchronously preparing the export and sending notification of export download available when complete.


jsavell (approved):

JamesS-TAMU:
Per Slack chats with @wwtamu , I tested ProQuest export with the `sprint-9-2048` branch and was able to export large zip packages successfully (I filtered by Graduation Semester which is what the graduate office uses, and tested `December 2023`, `August 2024`, which were previously failing), the issues I [commented here](https://github.com/TexasDigitalLibrary/Vireo/issues/2048#issuecomment-2813606585) no longer present. The zip exports were about 3GB+ and 4GB+ in size.

cstarcher (approved):

PR #2055, opened 2025-04-22
Issue #2048: Stream batch response

wwtamu:
These are not fully streaming as the database result set is still being stored in memory as an `ArrayList`. The submission repository will require a streaming batch dynamic submission query to minimize the minimal required memory for large batches.

This also presents an issue with the current error handling approach in which handles exception after writing headers and content to the output stream.

Previous to commit [854886ea28169c69f22d82ba8fa4a88cb7dd2eab](https://github.com/TexasDigitalLibrary/Vireo/pull/2055/commits/854886ea28169c69f22d82ba8fa4a88cb7dd2eab) when an exception occurred, it would append the `ApiResponse` to the stream and not provide any feedback and will result in a corrupt or incomplete file download.

With the latest changes to attempt to close all streams properly, set response headers before streaming content, and using a consistent pattern of close/operate on stream manifested a new regression with the @wvr/core client-side file download processing. The output streams closed properly disallowed the `ApiResponse` JSON to be written to the client.

I feel like the correct approach to a batch export file download is to determine the Content-Length to send in the response header before streaming the response. This affords the simplest client-side processing to determine initial success conditions. If the processing is done before streaming the response the status code can be correctly provided.

This requires several requirements to be defined:

1. Path strategy for storing batch export artifacts and their duration
2. Client timeout if batch export requires processing
3. Action email workflow rule to notify of available download


wwtamu:
Closing in favor of https://github.com/TexasDigitalLibrary/Vireo/pull/2061

Issue #2065, opened 2025-05-14
Event Time column not exporting correctly for Excel export.


jameswsullivan:
Describe the bug:

In [Issue #1995](https://github.com/TexasDigitalLibrary/Vireo/issues/1995) I fixed the Last Event column by accounting for the `ActionLog` object type, but it turned out that more than one column references the ActionLog type and we need to differentiate what value gets returned by the column title.

Solution:
Instead of using `value = actionLog.getEntry().toString();`, add a switch statement to differentiate by column name.

PR #2066, opened 2025-05-14
Issue #2065: Event Time column not exporting correctly for Excel export.

jameswsullivan:
Describe the bug:

In [Issue #1995](https://github.com/TexasDigitalLibrary/Vireo/issues/1995) I fixed the Last Event column by accounting for the `ActionLog` object type, but it turned out that more than one column references the ActionLog type and we need to differentiate what value gets returned by the column title.

Solution:
Instead of using `value = actionLog.getEntry().toString();`, add a switch statement to differentiate by column name.

### DIFF
--- a/src/main/java/org/tdl/vireo/model/packager/ExcelPackager.java
+++ b/src/main/java/org/tdl/vireo/model/packager/ExcelPackager.java
@@ -85,7 +85,7 @@ public class ExcelPackager extends AbstractPackager<ExcelExportPackage> {
                         }
                         Object valueAsObject = EntityUtility.getValueFromPath(submission, valuePath);
 
-                        String value;
+                        String value = "";
 
                         if (valueAsObject instanceof Calendar) {
                             Calendar calendar = (Calendar) valueAsObject;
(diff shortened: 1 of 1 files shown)


jameswsullivan:
Resolves: https://github.com/TexasDigitalLibrary/Vireo/issues/2065

wwtamu (approved):
Looking at the submission list column defined in the system defaults; this condition is necessary to provide the correct value when the value path returns an ActionLog.

Could you please add comments where these string literals are from, [SYSTEM_Default_Submission_List_Columns](https://github.com/TexasDigitalLibrary/Vireo/blob/main/src/main/resources/submission_list_columns/SYSTEM_Default_Submission_List_Columns.json#L162), and double check they are not mutable from the UI. Thank you.

JamesS-TAMU:
The string literals that are used to differentiate the column names are from:
- Last Event column: https://github.com/TexasDigitalLibrary/Vireo/blob/main/src/main/resources/submission_list_columns/SYSTEM_Default_Submission_List_Columns.json#L162
- Event Time colum: https://github.com/TexasDigitalLibrary/Vireo/blob/main/src/main/resources/submission_list_columns/SYSTEM_Default_Submission_List_Columns.json#L174

I poked around in `Custom columns` and settings and I don't believe the column names are mutable in the UI.

image

Issue #2079, opened 2025-06-06
file upload renaming does not change name to submission type


jrmartino:
**Describe the bug**
When making a submission, we'd like the PRIMARY document to be renamed according to the VUG recommendation of LASTNAME-SUBMISSIONTYPE-YEAR. Instead, we get LASTNAME-PRIMARY-YEAR.

**To Reproduce**
1. keep default app.documentType.rename as PRIMARY
2. login as Jack Daniels
3. enable submissions somewhere
4. start a submission
5. set a Submission Type to Dissertation
6. Upload a file**Expected behavior**
We expect to see the uploaded file renamed DANIELS-DISSERTATION-2025.pdf
**Screenshots**
I

![Image](https://github.com/user-attachments/assets/49b32d75-4f58-433a-a905-3b908c97b2b5)

**Additional context**
The problem is in the Submission controller, which sets the documentType to the type to be renamed instead of taking the value from the submission form.

jrmartino:
I have a fix here locally, will submit a PR

PR #2080, opened 2025-06-06
Fix uploadFile in SubmissionController so that it correctly renames uploaded files

jrmartino:
#2079 

This is a change from current behavior which just renames a PRIMARY file to LASTNAME-PRIMARY-YEAR
Adjusted behavior will give things like LASTNAME-DISSERTATION-YEAR etc.

### DIFF
--- a/src/main/java/org/tdl/vireo/controller/SubmissionController.java
+++ b/src/main/java/org/tdl/vireo/controller/SubmissionController.java
@@ -1058,8 +1058,12 @@ public class SubmissionController {
 
         if (documentTypesToRename.contains(documentType)) {
             String lastName = submission.getSubmitter().getLastName().toUpperCase();
+            String newType = "DOCUMENT"; // default
+            if (submission.getFieldValuesByPredicateValue("submission_type").size()>0 ) { //should be exactly 1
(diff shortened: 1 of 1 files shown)


wwtamu:
This is a reasonable feature adjustment with only questionable difference being the default file naming when submission type has not yet been selected.

The default naming before this change `LASTNAME-PRIMARY-YEAR` as described above. Now the naming respects the selection of submission type under document information otherwise defaulting to `LASTNAME-DOCUMENT-YEAR`. I believe this is a convention change that requires community consensus. 

I agree with the desired feature change but observe this brings in a potential for inconsistency that a student alone could have occur. If the student uploads a file before selecting a submission type or changing the submission type after uploading a file there will be an inconsistency with the metadata field value for submission type and the named file expecting it to match by the defined naming convention. This may be considered insignificant as long as administrators are aware of it.

Otherwise, should this change include synchronizing the filename, if any primary document uploaded, when changing the submission type?

jrmartino:
Good points. There are certainly a few things to think about. What if multiple types are to be rewritten (PRIMARY, LICENSE ...? The logic here needs to be restricted to just PRIMARY. Not sure about the file synchronization question, it would be interesting to look into. Is it possible to have multiple submission types on a submission? If so, to be careful we we'd only rename if there were exactly one.  Will push some changes onto PR.

PR #2112, opened 2025-10-16
Issue #2097: Enable HTML support for email templates.

### LINKED ISSUE #2097: Enable HTML support for email templates.
Recently we got a question about being able to add "links over text" in the email templates and making the emails accessibility compliant so that the screen readers can read the links' texts out loud, rather than the full URL, such as, use
`<a href="https://myurl.com/12345abcde">myurl</a>` , rather than just `https://myurl.com/12345abcde` appearing in the email.

@wwtamu quickly found the two lines in `VireoEmailSender.java` and I tested with our instance and it worked alright, only thing I needed to do was edit the email template's texts and added some `<p>`, `<br>`, `<a>` to them to make the formatting look correct.

Here is the modification: https://github.com/jameswsullivan/Vireo/tree/enable-mime-html-email

Are there any objections to enabling this by default, or potentially making it "configurable" in case there are cases that would require "plain text only"?

jameswsullivan:
Resolves [Issue 2097](https://github.com/TexasDigitalLibrary/Vireo/issues/2097).

I'm a bit hesitant on this one though because it probably requires the admins and users to review and edit all of their existing email templates, as some formatting might not appear correctly without the HTML tags.

### DIFF
--- a/src/main/java/org/tdl/vireo/service/VireoEmailSender.java
+++ b/src/main/java/org/tdl/vireo/service/VireoEmailSender.java
@@ -76,7 +76,7 @@ public class VireoEmailSender extends JavaMailSenderImpl implements EmailSender 
         loadConfiguration();
 
         MimeMessage message = createMimeMessage();
-        MimeMessageHelper mm = new MimeMessageHelper(message);
+        MimeMessageHelper mm = new MimeMessageHelper(message, true);
 
         mm.setFrom(vireoEmailConfig.getFrom());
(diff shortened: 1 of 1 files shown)


jameswsullivan:
Yep, that was what I was worried about too. I did this only as a test at first to see if it was possible to do "link over text" in an email template.
I'm guessing since this is not quite urgent, might table this for a future version or something, along with other potential accessibility changes?

cstarcher:
Yes, I'm not opposed to adding this functionality in a future release. I just want to make sure it doesn't break current functionality.

PR #2129, opened 2026-03-02
Issue 2128: Fix Committee Contact Email in Excel Export

### LINKED ISSUE #2128: Committee Contact Email column not exported in ExcelPackager
#### Bug Description

When exporting to Excel, the Committee Contact Email doesn't show in the export, with the following warning in the logs:

```
WARN 1 --- [io-9000-exec-16] o.t.vireo.model.packager.ExcelPackager   : Unable to get value from committeeContactEmail
```

#### To Reproduce

List view -> Customize columns -> enable Committee Contact Email column -> do an Excel export.

The email addresses are shown in the UI, but when doing a Download export/Batch Export in Excel the exported column is empty.

jameswsullivan:
Fix https://github.com/TexasDigitalLibrary/Vireo/issues/2128

- add a backing field in `committeeContactEmail` in `Submission.java`
- modify `getCommitteeContactEmail()` to return the value
- call `getCommitteeContactEmail()` in `ExcelPackager.java` before the value lookup so the value is getting generated
- `committeeContactEmail` should then be covered by the `else ... value = valueAsObject.toString();` clause in the export

I've tested on our instance.

### DIFF
--- a/src/main/java/org/tdl/vireo/model/Submission.java
+++ b/src/main/java/org/tdl/vireo/model/Submission.java
@@ -253,6 +253,9 @@ public class Submission extends ValidatingBaseEntity {
     @JsonView(Views.SubmissionList.class)
     private Map<Long, String> columnValues;
 
+    @Transient
+    private String committeeContactEmail;
+
     public Submission() {
         setModelValidator(new SubmissionValidator());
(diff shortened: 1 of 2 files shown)


cstarcher (approved):
Looks good. Thanks James!

PR #2138, opened 2026-03-25
Add degree level to DSpaceMets export

jsavell:
We noticed that the degree level metadata was missing when publishing to DSpace via DSpaceMets.

This PR adds the degree level metadata to match the behavior in V3.


cstarcher:
Resolves #2140

cstarcher (approved):

Issue #1699, opened 2023-04-14
Provide alternate login path for local authentication for when local authentication is disabled.


kaladay:
This feature is for assisting in supporting the feature from issue #1585 and the feature from #1593 in conjunction.

This is not a request to provide an alternate e-mail account.
This is a request to provide an alternate path for when the login and registrations modal and links are hidden.

This should provide a new option for `localAuthentication` like `localAuthentication="alternate"` such that it acts as if `localAuthentication=false` except an alternate login endpoint is presented.

The alternate login endpoint should provide the login modal when using `localAuthentication="alternate"`.

Issue #1593, opened 2023-03-03
Hide email/password login (optionally) when Shibboleth is configured (4.x Login)


nicksjunkyard:
How do we go about disabling vireo based login or make Shibboleth login the default? It would be nice to get rid of the register menu item and will be confusing to have two login choices. Thank you!

jcreel:
May relate to #1585

cstarcher:
Need to update README with instructions for removing non-Shibboleth login.

jcreel:
Add a configuration value (appConfig.js to avoid DB migration) to indicate if the "Login" link should go straight to Shibboleth or should bring up the modal.

Issue #1585, opened 2022-11-09
alternate email sign in on shibboleth-enabled instances


smutniak:
"The historical context would be to allow a former student (whose Shibboleth login credentials no longer exist) to login using an alternate email/password in case edits need to be made after they’ve gone. We had that issue come up several times during the Vireo 3 years, so it’s always possible it come up again."

While we do need a tech support sign in option I am also adding the above for discussion.

jcreel:
This has not been supported in previous versions of Vireo.

jcreel:
May relate to #1593

smutniak:
My apologies for leading with the student sign in issue,  more importantly the tech support issue is what is core.  The interest is in getting a Vireo4 counterpart to vireo3's /PasswordAuthentication as in https://utexas-etd.tdl.org/login/PasswordAuthentication that can be used to sign in to TDL hosted sites that use shibboleth authentication.  Currently I need to make a local replica to a site on a local VM to debug issues rather than signing in via the UI to the site (e.g. I don't have TTU, WTAMU, UTD, etc shibboleth credentials) so I can see what the user/admin sees in the UI.

cstarcher:
Password login not working behind proxy. Need to investigate how this works with DSpace.

PR #2142, opened 2026-04-24
Issue #2141: Add "signup" route for email verification when local login and register are hidden.

### LINKED ISSUE #2141: No path to bring up "Verify Email" modal when "localAuthentication" is set to "alternate"
**Describe the bug**
Currently when `localAuthentication` is set to `'alternate'`, we can bring up the local login modal from the `/login` page, but since the `Register` link in the upper right corner is hidden, I don't think there is currently a way to make `Verify Email` pop up. I was able to make it show up with
```
$('#verifyEmailModal').modal('show');
$('#verifyEmailModal').modal('hide');

angular.element('#verifyEmailModal').modal('show');
angular.element('#verifyEmailModal').modal('hide');
```
because it exists in the DOM but hidden, but this seems to be too much of a hacking way to do it.

**To Reproduce**
Set
```
LOCAL_AUTHENTICATION = 'alternate'
APP_USENETIDASIDENTIFIER = false
```
and then attempt to register a local account and login.

**Expected behavior**
When registering and logging in with a local account is not disabled but just the hidden, there should be a way to initiate the registration/verify email process too, similar to how the `/login` route was implemented.

jameswsullivan:
Resolves https://github.com/TexasDigitalLibrary/Vireo/issues/2141 .

I basically followed how @kaladay did with the `/login` route and the path with the least resistance seems to be just add a route/page to bring up the existing Verify Email modal that's hidden, otherwise if I choose to alter how the `/register` path behaves, it seems that I'd have to involve changing weaver stuff (`RegistrationController`?) and/or add conditional behaviors to separate it with the `/register?token=` stuff, which I'm not familiar with.

Related discussions:
- https://github.com/TexasDigitalLibrary/Vireo/issues/1699
- https://github.com/TexasDigitalLibrary/Vireo/pull/1700
- https://github.com/TexasDigitalLibrary/Vireo/pull/1697
- https://github.com/TexasDigitalLibrary/Vireo/issues/1585
- https://github.com/TexasDigitalLibrary/Vireo/issues/1593

I tested with our instance and it worked for me, but more tests on other instances are always helpful.

### DIFF
--- a/build/appConfig.js.template
+++ b/build/appConfig.js.template
@@ -38,7 +38,7 @@ var appConfig = {
       Set this to true or false booleans or set it to the literal strings (with quotes) 'alternate' or 'alternateAllowRegister'.
       When set to the boolean true, the login modal and registrations links are available.
       When set to the boolean false, a modal is not used and Shibboleth login is automatically attempted and no registration link is shown.
-      When set to the string 'alternate' to hide the local login (and registration) as if localAuthentication is false and provide an alterate (hidden) login page for local authentication.
+      When set to the string 'alternate' to hide the local login (and registration) as if localAuthentication is false and provide an alternate (hidden) login page for local authentication.
       When set to the string 'alternateAllowRegister' to hide the local login the same way as 'alternate', but allow for account registration at the /signup route.
      */
     'localAuthentication': ${LOCAL_AUTHENTICATION},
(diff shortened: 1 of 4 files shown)


jameswsullivan:
- Add a `localAuthentication = 'alternateAllowRegister'` option in `appConfig` to carry out this function to avoid altering the current `'alternate'`'s behaviors.
- Hide the sign-up option when `localAuthentication` is set to `'alternate'` or `false`.

cstarcher (approved):

PR #2151, opened 2026-05-18
Enable HTML support for email templates

### LINKED ISSUE #2097: Enable HTML support for email templates.

jsavell:
Resolves #2097

Continuation of #2112

Includes the original commits from 2112 to enable HTML emailing, while also updating default templates with anchor tags for links.

We can make further changes to the templates if needed via this PR.


cstarcher:
@jsavell 
My only concern about this is that it breaks the formatting for the default emails. This is the same concern I had with PR #2112 from @jameswsullivan.

This is the original formatting.

Screenshot 2026-07-03 at 9 42 09 AM

This is the formatting with these changes.

Screenshot 2026-07-03 at 9 42 24 AM

jameswsullivan:
Right, at the time when I tested, I went in and manually edited the templates to add `<br>` and `<p>`s to keep the formatting from all collapsing into one. Maybe we should add `<br>`s as well to where there used to be line breaks?

cstarcher:
@jameswsullivan 
That is an acceptable solution to me. My only requirement is that we don't break the default formatting.

PR #2153, opened 2026-05-19
added loadActiveFilters function to retain sort order

### LINKED ISSUE #2165: name sorting not persistent in list upon reload
From comment in ticket #1959

I have enabled a sorting option (Last Name Ascending) when I am working in Vireo. I select an individual submission to work on it, and when I am finished, I select the List tab to get back to view the submissions. However, the sort preference I enabled does not exist and the original default sort (ID) is active.

Doing a page refresh doesn't really do anything, but I will usually click on the logo to go back to the student submission start page and then select the "List" from the Admin drop-down menu. At that point, my sort preference reappears and is active.

I frequently need to repeat those steps during constant work to ensure that that the sort preference stays active, but this is clearly not ideal.

cstarcher:
## Description
Resolves #2165 

## Type of Change
- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

## AI Usage Disclosure

**AI Tool(s) Used Including Version:**
ChatGPT 5.4 mini

**Nature of Use:**
  - [X] Code generation
  - [X] Code suggestions/refactoring
  - [ ] Documentation
  - [ ] Tests
  - [ ] Other (please describe)

**Additional Description:**
Used to identify and solve the problem.

- [ ] No AI Used


smutniak:
This does maintain sort order but the original issue also mentions the problem of zero entries until a refresh.  Once the refresh happens the sort order is persistent but still needs a refresh.  I'm ok with this fix but we should create a new ticket for the original problem

smutniak (dismissed):

